Issue #3, opened 2023-02-13
Create initial API views


lemontree210:
The views should be based on current knowledge of business logic and can be extended in future iterations.

lemontree210:
Closing as too general

PR #9, opened 2023-02-24
Endpoint: count of users

aakhterov:
1. Common structure of pacakges 
2. Endpoint for count of users


lemontree210:
@aakhterov please merge `master` into your branch to resolve conflicts

api/serializers/people_serializers.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework import serializers
+from api.models import people
+
+
+class PersonalInfoSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = people.PersonalInfo
+        fields = ['id', 'date_and_time_added', 'first_name', 'last_name', 'tg_username',
+                  'email', 'phone', 'tz_summer_relative_to_utc', 'tz_winter_relative_to_utc',
+                  'age', 'information_source', 'native_languages', 'availability_slots',
```
lemontree210, line 10: some of these fields changed in `PersonalInfo`

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
...
+        self.email = email
+        self.first_name = first_name
+        self.last_name = last_name
+
+    @classmethod
+    def from_dict(cls, input_dict):
+        # Check input parameters.
+        invalid_request = InvalidRequestObject()
+        # https://core.telegram.org/bots/api#chat : chat_id - max 52 significant bits
+        # 9007199254740991 is the max number you can store in the 53 bits => len(chat_id) <= 16
```
lemontree210, line 26:
```suggestion
        # 9007199254740991 is the max number you can store in the 53 bits. Hence len(chat_id) <= 16
```
It looks a bit confusing otherwise :)

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
...
+        if input_dict.get("registration_bot_chat_id") and \
+                len(input_dict.get("registration_bot_chat_id")) > 16:
+            invalid_request.add_error("registration_bot_chat_id", "Invalid length")
```
lemontree210, line 29: Did `black` approve this (`if .... and \`)? I suspect you didn't run pre-commit hooks :)

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if input_dict.get("tg_username") and (len(input_dict.get("tg_username")) < 5 or
+                                              len(input_dict.get("tg_username")) > 32):
```
lemontree210, line 32: same. I don't think `black` would like this :) I suggest we use it to keep our coding style consistent.

Same remark re: constants. It's best to keep "magic numbers" out of the code.

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
...
+        self.last_name = last_name
+
+    @classmethod
+    def from_dict(cls, input_dict):
+        # Check input parameters.
+        invalid_request = InvalidRequestObject()
+        # https://core.telegram.org/bots/api#chat : chat_id - max 52 significant bits
+        # 9007199254740991 is the max number you can store in the 53 bits => len(chat_id) <= 16
+        if input_dict.get("registration_bot_chat_id") and \
+                len(input_dict.get("registration_bot_chat_id")) > 16:
```
lemontree210, line 28: I still suggest extracting things like `16` into constants

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
...
+            invalid_request.add_error("registration_bot_chat_id", "Invalid length")
+
+        if input_dict.get("tg_username") and (len(input_dict.get("tg_username")) < 5 or
+                                              len(input_dict.get("tg_username")) > 32):
+            invalid_request.add_error("tg_username", "Invalid length")
+
+        # See modal PersonalInfo
+        if input_dict.get("first_name") and len(input_dict.get("first_name")) > 100:
```
lemontree210, line 36: same here and later: here it's even more obvious because "100" repeats multiple times.

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
...
+        # 9007199254740991 is the max number you can store in the 53 bits => len(chat_id) <= 16
+        if input_dict.get("registration_bot_chat_id") and \
+                len(input_dict.get("registration_bot_chat_id")) > 16:
+            invalid_request.add_error("registration_bot_chat_id", "Invalid length")
+
+        if input_dict.get("tg_username") and (len(input_dict.get("tg_username")) < 5 or
+                                              len(input_dict.get("tg_username")) > 32):
+            invalid_request.add_error("tg_username", "Invalid length")
+
+        # See modal PersonalInfo
```
lemontree210, line 35:
```suggestion
        # See model PersonalInfo
```

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
...
+        if input_dict.get("first_name") and len(input_dict.get("first_name")) > 100:
+            invalid_request.add_error("first_name", "Invalid length")
+
+        # See modal PersonalInfo
```
lemontree210, line 39:
```suggestion
        # See model PersonalInfo
```

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
...
+        if input_dict.get("last_name") and len(input_dict.get("last_name")) > 100:
```
lemontree210, line 40: BTW maybe strings should be extracted into constants as well, but I'm not sure. Numbers definitely, though.

api/share/responses.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @classmethod
+    def build_from_none_request(cls, use_case_name: str):
+        errors = [f"No request in {use_case_name}"]
+        return cls(errors)
+
+    @classmethod
+    def build_from_error(cls, object_name: str, exc: Exception):
+        errors = [f"{str(exc)} in {object_name}"]
+        return cls(errors)
```
lemontree210, line 37: add blank line at EOF :)

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
...
+        if input_dict.get("email") and len(input_dict.get("email")) > 100:
+            invalid_request.add_error("email", "Invalid length")
+
+        # Phone must start with + or 00
+        if input_dict.get("phone") and not re.match(r'^\+|00', input_dict.get("phone")):
+            invalid_request.add_error("phone", "Incorrect format")
+
+        if input_dict.get("phone") and len(input_dict.get("phone")) > 12:
+            invalid_request.add_error("phone", "Invalid length")
+
```
lemontree210, line 55: you might want to check the new `utc_timedelta` param as well, and `availability_slots`

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
+import re
+from api.share.requests import ValidRequestObject, InvalidRequestObject
+
+
+class GetUsersCountRequest(ValidRequestObject):
+
+    def __init__(self,
+                 registration_bot_chat_id: str,
```
lemontree210, line 8: this doesn't look like it conforms with `black`...

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
+import re
+from api.share.requests import ValidRequestObject, InvalidRequestObject
+
+
+class GetUsersCountRequest(ValidRequestObject):
+
```
lemontree210, line 6: docstring?

api/share/tools.py
```diff
@@ -0,0 +1,28 @@
...
+            "self": current_url
+        }
+    }
+
+    if isinstance(data, dict):
+        http_response["_links"]["collection"] = "/".join(current_url.split("/")[:-2])
+    elif isinstance(data, list):
+        pass
+    else:
+        raise ValueError("data must be dict or list")
```
lemontree210, line 26: `TypeError`?

api/share/tools.py
```diff
@@ -0,0 +1,28 @@
+from typing import Dict, List, Union
```
lemontree210, line 1: I require Python 3.9 (or even 3.10) or higher, so you can leave just the `Union`.

api/share/tools.py
```diff
@@ -0,0 +1,28 @@
...
+from http import HTTPStatus
+
+
+def build_correct_http_response(data: Union[Dict, List],
```
lemontree210, line 5:
```suggestion
def build_correct_http_response(data: Union[dict, list],
```
See my note above on Python version.
It could also be useful to indicate what the dict/list contain (e.g. `list[str]`).
Do you also accept tuple? Maybe use `Sequence`?

api/share/tools.py
```diff
@@ -0,0 +1,28 @@
...
+                                current_url: str,
+                                status: HTTPStatus,
+                                errors: List = None
```
lemontree210, line 8:
```suggestion
                                errors: list = None
```
`list[something]` would be nice

api/share/tools.py
```diff
@@ -0,0 +1,28 @@
...
+            "statusCode": status.value,
+            "messages": errors if errors else {"status": status.phrase}
+        },
+        "_links": {
+            "self": current_url
+        }
+    }
+
+    if isinstance(data, dict):
+        http_response["_links"]["collection"] = "/".join(current_url.split("/")[:-2])
```
lemontree210, line 22: a CSI comment would be nice here. I don't think I understand the intention, and where [:-2] comes from

api/share/tools.py
```diff
@@ -0,0 +1,28 @@
...
+    elif isinstance(data, list):
+        pass
+    else:
+        raise ValueError("data must be dict or list")
+
+    return http_response
```
lemontree210, line 28: blank line at EOF :)

api/use_cases/__init__.py
```diff
@@ -0,0 +1,17 @@
+from api.share.responses import ResponseFailure, ResponseSuccess, Response
+from api.share.requests import ValidRequestObject
+
+
+class UseCase:
+    def execute(self, request_object: ValidRequestObject) -> Response:
```
lemontree210, line 6: docstring? When I see this the first time, I don't understand what it's here for

api/use_cases/get_users_count.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def process_request(self, request_object: GetUsersCountRequest):
+        # business logic
+        registration_bot_chat_id = request_object.registration_bot_chat_id
+        tg_username = request_object.tg_username
+        phone = request_object.phone
+        email = request_object.email
+        first_name = request_object.first_name
+        last_name = request_object.last_name
+
```
lemontree210, line 20: utc_timedelta, availability_slots, communication_language_mode, comment

api/use_cases/get_users_count.py
```diff
@@ -0,0 +1,35 @@
...
+        if tg_username:
+            res = res.filter(tg_username__iexact=tg_username)
+        if phone:
+            res = res.filter(phone=phone)
+        if email:
+            res = res.filter(email__iexact=email)
+        if first_name:
+            res = res.filter(first_name__iexact=first_name)
+        if last_name:
+            res = res.filter(last_name__iexact=last_name)
```
lemontree210, line 33: it looks like this workflow will lead to too many database queries. Or are they queried once (all of them are SELECTed) and then stored in memory by the time you run `filter`? In this case, when the amount of `PersonalInfo` object gets into tens of thousands, we may be in a little trouble?

api/views/people.py
```diff
@@ -0,0 +1,44 @@
...
+from api.use_cases.get_users_count import UsersCount
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from http import HTTPStatus
+
+
+class UserCountView(APIView):
+
+    def get(self, request):
+        try:
```
lemontree210, line 13: this is way too many statements inside a `try` block. Ideally it should contain only one statement

api/views/people.py
```diff
@@ -0,0 +1,44 @@
...
+            params = request.query_params
+            request_object = GetUsersCountRequest.from_dict(params) # object contains correct parameters
+            # if something went wrong
```
lemontree210, line 16: what's more, you have "if something went wrong" inside a `try` block. That is another code smell to me

api/share/responses.py
```diff
@@ -0,0 +1,37 @@
+from abc import ABC
```
lemontree210, line 1: maybe this file should also be in a separate package for symmetry?

lemontree210 (changes requested):
Thanks for the PR! I have 3 general questions/requests.

1. Could you add a README for `requests` and maybe `responses` (if you put it in a separate package) to explain the logic? I understand that this is familiar workflow for you and it seems self-explanatory, but I sometimes find it hard to understand your approach, and maybe I will not be the only one :)
2. Please install pre-commit hooks, I really doubt that you ran `black` on the codebase. Maybe I'm wrong, but I'm pretty sure it wouldn't let you have some indents/brackets/slashes I'm seeing here.
3. Is this "user count" view really a view that will be used in our project? I thought it was more like an exercise. I'm not sure I can right now see the use case for it.
4. (Just duplicating my message from earlier) I think something went wrong with merging `master`. Some of the files shouldn't be marked as "files changed", as far as I understand.

aakhterov:
_1. Could you add a README for requests and maybe responses (if you put it in a separate package) to explain the logic? I understand that this is familiar workflow for you and it seems self-explanatory, but I sometimes find it hard to understand your approach, and maybe I will not be the only one :)_
Я думаю, что лучше объяснить такую логику в основном README. Я собираюсь это сделать (см. TODO в основном README)

_2. Please install pre-commit hooks, I really doubt that you ran black on the codebase. Maybe I'm wrong, but I'm pretty sure it wouldn't let you have some indents/brackets/slashes I'm seeing here._
Yes, my bad. I haven't installed pre-commit hooks and black. I'll fix it.

_3. Is this "user count" view really a view that will be used in our project? I thought it was more like an exercise. I'm not sure I can right now see the use case for it._
We have two requests from bot: (1) if user with given registration_bot_chat_id exists (2)  if user with given first name, second name and etc. exists. This endpoint returns users count with giver parameters: if 0 - there are no such users, if not 0 users exist.

_4. (Just duplicating my message from earlier) I think something went wrong with merging master. Some of the files shouldn't be marked as "files changed", as far as I understand._
Maybe there are some mistakes. I'll see

README.md
```diff
@@ -1,2 +1,23 @@
...
+5. Create database migrations: `python manage.py makemigrations api`
+6. Apply database migrations: `python manage.py migrate`
+7. Create a superuser for admin interface: `python manage.py createsuperuser`
+8. Run the server locally: `python manage.py runserver`
+
+The main page is the admin page, as no user-facing web interface is planned yet.
+
+### Some information about an architecture 
+
+1. Django implements MTV (Model-Template-View) architecture pattern. This is almost MVC (Model-View-Controller) pattern, but with a few changes. Model is Model, Template is View and View is Controller (https://medium.com/shecodeafrica/understanding-the-mvc-pattern-in-django-edda05b9f43f).
```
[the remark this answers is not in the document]
aakhterov: It looks like comment for comment :)

api/use_cases/get_users_count.py
aakhterov: As a understand, django's queries are lazy. "QuerySets are lazy – the act of creating a QuerySet doesn’t involve any database activity. You can stack filters together all day long, and Django won’t actually run the query until the QuerySet is evaluated." In this case SQL query evaluated in res.count() and it looks like "SELECT ..... WHERE () AND () AND () ...."

api/views/people.py
aakhterov: Yes, you're right. I think to remove this _try .. expect_ at all but later. It's just in case now. In principle, _except_ part of this block doesn't have to get control at all.

api/views/people.py
aakhterov: This is comment for the next line of code: _if isinstance(request_object, InvalidRequestObject):_. This construction allows to handle errors inside user cases and not raise them to this level. In particular because of this we can remove try..expected block (see previous comment).

api/share/responses.py
aakhterov: It may not be suitable for this. Because we'll have a number of requsets (one requset corresponds to one use case) and only two responses - ResponseFailure and ResponseSuccess, which correspond to success or fail in user cases.

api/serializers/people_serializers.py
aakhterov: I agree

api/share/requests/users.py
aakhterov: Yes, my fault. I didn't use _black_

api/use_cases/get_users_count.py
aakhterov: What do you mean?

README.md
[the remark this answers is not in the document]
lemontree210: it's a markdown hyperlink. The text "this article" will become a hyperlink.

api/use_cases/get_users_count.py
lemontree210: I mean that these are new attributes that we now have in `PersonalInfo`

api/use_cases/get_users_count.py
lemontree210: OK great!

api/views/people.py
lemontree210: I know, but it looks like you put something into `try` and then if something goes wrong you don't call `except` but continue within `try`. That's just how it looks for an outside reader :)

api/views/people.py
lemontree210: I just think that before we merge this into `master`, we should make sure that `try` only tries one statement. We can extract that block into a function if no other changes are possible.

api/share/responses.py
lemontree210: OK got it

lemontree210:
> We have two requests from bot: (1) if user with given registration_bot_chat_id exists (2) if user with given first name, second name and etc. exists. This endpoint returns users count with giver parameters: if 0 - there are no such users, if not 0 users exist.

OK, makes sense, thank you.

lemontree210:
> We have two requests from bot: (1) if user with given registration_bot_chat_id exists (2) if user with given first name, second name and etc. exists. This endpoint returns users count with giver parameters: if 0 - there are no such users, if not 0 users exist.

I was just thinking we could use [get_object_or_404()](https://docs.djangoproject.com/en/4.1/topics/http/shortcuts/#get-object-or-404) for this, without counting. It's like we don't really need the quantity, we need "0" or "not 0". It sounds like we're doing something extra that the bot doesn't ask to do.

aakhterov:
> > We have two requests from bot: (1) if user with given registration_bot_chat_id exists (2) if user with given first name, second name and etc. exists. This endpoint returns users count with giver parameters: if 0 - there are no such users, if not 0 users exist.
> 
> I was just thinking we could use [get_object_or_404()](https://docs.djangoproject.com/en/4.1/topics/http/shortcuts/#get-object-or-404) for this, without counting. It's like we don't really need the quantity, we need "0" or "not 0". It sounds like we're doing something extra that the bot doesn't ask to do.

I think 404 response is not a good idea for our purposes. Imho 404 response is an error. In our case if user doesn't exist this isn't an error, this is a normal situation. Therefore I response 200 in both cases. Futher, we don't need to return an entire object to these requests. We could make these requests look like ..../users/exist?par1=.... , but in this case we have to return boolean value and I don't like it much.

api/share/requests/users.py
aakhterov: > same. I don't think `black` would like this :) I suggest we use it to keep our coding style consistent.
> 
> Same remark re: constants. It's best to keep "magic numbers" out of the code.

Do we suppose these numbers will change? If no and if we don't use them somewhere else and if we have comments about these values, what is the benefit to keep them in constants?

api/share/requests/users.py
aakhterov: as previous comment

api/share/requests/users.py
aakhterov: Ok, I can agree with this

api/share/requests/users.py
aakhterov: I think such questions are good for refactoring stage

api/share/responses.py
aakhterov: Ok

api/share/requests/users.py
aakhterov: Why? We want to get users count by utc_timedelta or availability_slots?

api/share/requests/users.py
aakhterov: Yes, I know

api/share/requests/users.py
aakhterov: Ok

api/share/tools.py
aakhterov: Ok. I agree

api/share/tools.py
aakhterov: Ok, I'll fix it. But I'm sure it isn't the most important things we need to do now. During refactoring - yes, but not now.

api/share/tools.py
aakhterov: see previous comment

api/share/tools.py
aakhterov: the same

api/share/tools.py
aakhterov: Ok

api/use_cases/get_users_count.py
aakhterov: Why do they need here?

api/views/people.py
aakhterov: I prefer to remove it.

api/views/people.py
aakhterov: See previous comment

api/use_cases/__init__.py
aakhterov: Ok

api/share/tools.py
aakhterov: Ok

README.md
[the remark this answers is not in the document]
aakhterov: Ok

api/share/requests/users.py
lemontree210: Well I don't know, it's just [good coding](https://refactoring.guru/replace-magic-number-with-symbolic-constant) style to me, especially when same number is repeated over and over again. I can understand magic numbers in tests, but not in working code, where they are repeated.

api/share/requests/users.py
lemontree210: Sure, but just keep in mind that `master` should receive code that is already refactored in this way :)

api/share/requests/users.py
lemontree210: Right, no, we don't. I thought you were checking the correctness of all essential params. For the count, we don't. But then will you be checking the correctness of `utc_timedelta` elsewhere? Maybe we should do all the checks in one place (not where the count is computed)?

api/share/tools.py
lemontree210: Absolutely, just see my comment above: I would like to merge things into `master` when questions about code are resolved.

api/use_cases/get_users_count.py
lemontree210: OK, I see your point.

lemontree210:
> I think 404 response is not a good idea for our purposes. Imho 404 response is an error. In our case if user doesn't exist this isn't an error, this is a normal situation.

@aakhterov Well that's the thing: if you **count** the users then yes, it is correct that `0` is a normal result and requires a 200 response. But the main purpose of the view is to tell the bot whether the user exists or not. So the most obvious (to me, and maybe to an outside observer) way to go is to try and request the user with the given params, just like you request a web page. And you either get it and get the content with 200 response, or you get no page or user and get a 404 response. To me this looks like a very standard and predictable scenario for a case of requesting one resource that may or may not exist. That's why I thought that counting is a bit confusing solution for the task we have here.

api/share/requests/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class ValidRequestObject(RequestObject):
+
+    @classmethod
+    def from_dict(cls, input_dict):
+        raise NotImplementedError
+
+    def __bool__(self):
+        return True
```
lemontree210, line 30: Is it really necessary? Any object is truthy. Even without this method the following will work:
```py
o = ValidRequestObject()
if o:
    do_something()
```
and `bool(o)` will also return `True`.

And, of course, even with this method `o is True` will return `False`.

api/share/requests/users.py
```diff
@@ -0,0 +1,65 @@
...
+        self.tg_username = tg_username
+        self.phone = phone
+        self.email = email
+        self.first_name = first_name
+        self.last_name = last_name
+
+    @classmethod
+    def from_dict(cls, input_dict):
+        # Check input parameters.
+        invalid_request = InvalidRequestObject()
```
lemontree210, line 24: I would love to have a comment explaining why we're instantiating an object for an invalid request here. Maybe make the comment "check input parameters" a bit more verbose. 

As an outside reader, I'm rather expecting something like a "RequestObject" that can become valid or invalid during validation. Not an `InvalidRequestObject`, for which we then check if it has errors. If I misunderstand the logic, the comment would help me.

api/share/requests/users.py
lemontree210: Just in case: putting a link to [section on validation](https://docs.djangoproject.com/en/4.1/ref/models/instances/#validating-objects) here, where built-in validation methods are mentioned.

api/share/responses.py
```diff
@@ -0,0 +1,37 @@
...
+    pass
+
+
+class ResponseSuccess(Response):
+
+    def __init__(self, value=None):
+        self.value = value
+
+    def __bool__(self):
+        return True
```
lemontree210, line 14: same as above. Maybe you're just doing this for clarity in contrast to `ResponseFailure`, then a comment is worth adding.

api/use_cases/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+        if not request_object:
+            return ResponseFailure.build_from_none_request(self.__class__.__name__)
```
lemontree210, line 8: actually, `if not request_object` doesn't necessarily mean that `request_object is None`, it's just falsy. So maybe we should either rename `build_from_none_request()` or put `if request_object is None`, depending on what's more accurate.

lemontree210 (changes requested):
Some other minor remarks I spotted during the second reading

api/share/requests/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class RequestObject(ABC):
+    pass
+
+
+class InvalidRequestObject(RequestObject):
+
+    def __init__(self):
+        self.errors = []
```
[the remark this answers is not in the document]
aakhterov: No, it's no internal variable. It uses in views

api/share/requests/__init__.py
aakhterov: It's needed for corresponding to the InvalidRequestObject, where it returns False

api/share/requests/users.py
aakhterov: Because InvalidRequestObject has a method to add errors. We have ValidRequestObject and InvalidRequestObject

api/serializers/people_serializers.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework import serializers
+from api.models import people
+
+
+class PersonalInfoSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = people.PersonalInfo
+        fields = ['id', 'date_and_time_added', 'first_name', 'last_name', 'tg_username',
+                  'email', 'phone', 'tz_summer_relative_to_utc', 'tz_winter_relative_to_utc',
```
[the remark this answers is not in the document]
aakhterov: > > I think 404 response is not a good idea for our purposes. Imho 404 response is an error. In our case if user doesn't exist this isn't an error, this is a normal situation.
> 
> @aakhterov Well that's the thing: if you **count** the users then yes, it is correct that `0` is a normal result and requires a 200 response. But the main purpose of the view is to tell the bot whether the user **exists or not**. So the most obvious (to me, and maybe to an outside observer) way to go is to try and request the user with the given params, just like you request a web page. And you either get it and get the content with 200 response, or you get no page or user and get a 404 response. To me this looks like a very standard and predictable scenario for a case of requesting one resource that may or may not exist. That's why I thought that counting is a bit confusing solution for the task we have here.

I got you logic. My logic is above. We can do as you describe, we can do as I've done. More over we can invent more variants and they also will be appropriate.  Also we can find pro and cons in all variants. If to be honest, I don't see a sense to talk about it endlessly.

api/share/responses.py
aakhterov: see previous comments

api/use_cases/__init__.py
aakhterov: I got it

aakhterov:
@lemontree210 , If you want to merge into master only ideal code, maybe will we create branch test or dev for working, but not perfect code?

api/serializers/people_serializers.py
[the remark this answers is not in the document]
lemontree210: You can do as you wish, as long as it works and as long as an outside developer understands why you did this to solve a problem of existence of an object.

api/share/requests/__init__.py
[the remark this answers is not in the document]
lemontree210: It's a bit strange then, that on the one hand you "protect" it with methods (in the calling code you don't check for `len(o.errors)` but rather `o.has_errors()`), and then in other places you use the attribute directly. It looks like inconsistency to me.

api/share/requests/__init__.py
lemontree210: Well you can leave a corresponding comment then, because to an outside observer that will be strange. To me, though, it's strange that one class "knows" about its _sibling's_ internal structure. 

If you need to point out that in some subclass the `__bool__` method is handled differently, I'd suggest defining that `__bool__` method in `RequestObject` instead of `ValidRequestObject`. It will still look a bit strange to me, but at least it will be clear (perhaps a comment can be added in the parent class) that you did it to give a signal that you override it in one of the subclasses (and an IDE such as PyCharm will show where exactly).

api/share/requests/users.py
lemontree210: Yes, I understand that. But my "simple" logic says that it's strange that you create an object for an invalid request even before you know whether the request is valid or not (just because that object has a necessary method - but this object is not legacy code, you made it that way). I would expect something like a `Request` object that can have an `.is_valid` attribute, for example. Or a `Request` object that has an `.errors` list (or set) attribute and an `.is_valid()` method that returns `True` if `.errors` is empty and `False` otherwise.

api/share/responses.py
lemontree210: yep. and I'm skeptical about this approach. A class shouldn't know anything about how another child of its parent class works.

lemontree210:
> If you want to merge into master only ideal code, maybe will we create branch test or dev for working, but not perfect code?

@aakhterov "Ideal" and "clean" are different things to me :) I do believe that `master` should have code that has been reasonably refactored and has tests that accompany the new features. I'm not sure a different branch structure is needed. You can just merge `master` in your branch if `master` is updated before your pull request is merged.

And then if we run code from a dev branch, we risk never really completing the code and leaving it there because it "just works anyway".

But I'll keep this idea in mind, we'll see how it goes. Maybe we'll be forced to do that.

api/serializers/people_serializers.py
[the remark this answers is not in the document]
aakhterov: > You can do as you wish, as long as it works and as long as an outside developer understands why you did this to solve a problem of existence of an object. Right now I'm sure they would have same questions as I do ("why count to check for existence").

Because, it exists in the documentation. The outside developer also maybe surprise, why do we return an entiry objects with some fields in response. And why does he or she get an error in response. It depends on developer.

api/share/requests/users.py
aakhterov: We want to return either this class `GetUsersCountRequest` or `InvalidRequestObject` with errors. We try to separete responosibility: `ValidRequestObject` is resposible for correct input for use cases and `InvalidRequestObject` is in resposible for errors.

api/share/responses.py
aakhterov: Don't understand. `ResponseSuccess` doesn't know anything about `ResponseFailure`

api/share/requests/users.py
aakhterov: Ok, I got it

api/share/requests/users.py
aakhterov: Ok, I got it

api/share/requests/users.py
aakhterov: Ok, I got it

api/share/requests/users.py
aakhterov: Ok, I got it

api/share/tools.py
aakhterov: Ok, I got it

api/share/tools.py
aakhterov: Ok, I got it

api/serializers/people_serializers.py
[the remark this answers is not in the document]
lemontree210: It actually depends on our task. Maybe the bot will want to show something informative to the user. Or the backend just gets the entire object, but returns only `True` or `False` to the bot.

At any case, we both agree that our task is to provide the bot with the information whether the user **exists**. At least we should make the backend return `True` or `False` instead of number; and change the url:

```
urlpatterns = [
    path('users/count/', people.UserCountView.as_view()),
]
```
The bot shouldn't need to send a request to an url with `/count/` when all it needs is a check for existence. What happens "under the hood" - you're absolutely right - is the backend's choice (counting, getting object or 404, whatever). But the bot must think that it is simply asking for existence.

api/share/requests/users.py
lemontree210: All right, I think I can see your point, just please make sure this approach is explained :) Like this reply you just posted can be well used in a comment.

api/share/responses.py
lemontree210: No but `ResponseSuccess` sort of knows that `ResponseFailure` evaluates to `False` when `bool(obj)` is called on it. Because otherwise it would not contain code that is not needed (for any practical reasons), since, as we both agree, `bool(obj)` will return `True` for it anyway without any additional methods.

I really don't see how putting a method that does nothing helps. I think probably the only use case for methods that do nothing is a parent abstract class.

lemontree210:
@aakhterov Before I forget: some tests are obviously needed for this view. Maybe I missed them.

lemontree210:
> We could make these requests look like ..../users/exist?par1=.... , but in this case we have to return boolean value and I don't like it much.

Why not? It reflects perfectly what the bot needs to find out.

api/share/responses.py
aakhterov: > since, as we both agree, `bool(obj)` will return `True` for it anyway without any additional methods.

Yes, but it may not be clear for an outside developer (as you like to talk :) )

api/serializers/people_serializers.py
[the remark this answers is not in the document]
aakhterov: Ok we are in a circle. I don't want to argue

api/share/requests/__init__.py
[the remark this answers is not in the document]
aakhterov: Ok, I got it.

api/share/requests/__init__.py
aakhterov: Ok, I got it.

api/share/requests/users.py
aakhterov: Ok, I got it.

api/serializers/people_serializers.py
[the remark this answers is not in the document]
aakhterov: @lemontree210 , I don't want to argue constantly. I waste more time for disputes then for coding. If you don't agree with my point of view - no problems - please, design endpoints, I'll implement them and we'll forget about this.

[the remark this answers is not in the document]
lemontree210: @aakhterov To me, that's not how teamwork works... We present logical arguments and come to a mutually agreeable solution. My argument is: the bot only need to know about the existence of an object. Why should it send a request to a `/count/` url? How does that enforce principles of loose coupling and other principles we both agree are important?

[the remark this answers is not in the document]
aakhterov: My logic is the following: We have 3 roughly identical options. 
**1) Return an entire object.** 
Pro: 
- unversality (can uses somewhere else)
Cons:
- it's not obviously which fields do we return (only from PersonalInfo or PersonalInfo + data from joined tabels)
- There are can be more then an one object with parameters from request. Hence we need to return list and add pagination in the response. And all this is unnecessary inforamtion for this request.  

**2) Return only boolean value**
Pro: 
- return only what we need in this request
Cons:
- return only that we need in this request (this option is also cons because it makes this endpoint very specific, we won't be able to use it somewhere else)
- in this note [https://criteria.sh/blog/returning-booleans-in-responses](https://criteria.sh/blog/returning-booleans-in-responses) talks about something similar. I understand that it isn't exactly our case, but they don't recommend to use boolean value in response. Instead of this they offer "change the Boolean value to an Enumeration".

**3) Return something in between** - not so big as list of objects, but not so specific as only boolean value. For example, objects count
Pro: 
- more universal than option #2
- no unnecessary data   
Cons:
- less obvious than option #2
- less universal than option #1 

> To me, that's not how teamwork works...

I've chosen option #3, but in general I think all this three options are identical. Hence, I don't want to spend so much time arguing about the decision that isn't fundamental. As for me, this isn't teamwork object. Espessialy when we have many other things to do. 

> Why should it send a request to a `/count/` url?

Because if you are an outside developer and you need to check user existing, you look at the API descriptions and you see an endpoint that returns users count and I think you'll understand you can use it for your purpose.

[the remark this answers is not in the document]
lemontree210: Thank you for the link on boolean values in REST APIs, I agree that simply returning `True` (and nothing else) in a response isn't a good idea.

PR #12, opened 2023-03-09
Models: data migration for pre-populated data, minor model change (including removal of `InternalModelWithName`), fixing/ignoring linter errors

lemontree210:
- Removes `InternalModelWithName` as it's only used in `Language` and no longer needed
- Removes `AgeRange` from admin interface
- Adds field `registration_bot_language` to `PersonalInfo`
- Fixes `ruff` errors (after options were added to `pyproject.toml` and adds ignore rule for `A003 Class attribute `...` is shadowing a python builtin`

Pre-populates:
- [x] `TimeSlot`
- [x] `DayTimeSlot`
- [x] `LanguageLevel`
- [x] `Language`
- [x] `LanguageAndLevel`
- [x] `AgeRange`

The constants defined in `constants.py` can change (and in case of age ranges for group matching, most likely will), but they are not the focus of this PR.


api/migrations/0002_data_migration.py
```diff
@@ -0,0 +1,112 @@
+import datetime
+
+from django.db import migrations
+from django.db.backends.sqlite3.schema import DatabaseSchemaEditor  # for typing only
```
Roxe322, line 4: You can remove these comments about  typing :)

api/models/age_ranges.py
```diff
@@ -3,21 +3,28 @@
 from api.models.constants import DEFAULT_CHOICE_CHAR_FIELD_MAX_LENGTH
 
 
+# defining outside of model to use in data migration
```
Roxe322, line 6: I think this comment also is not necessary, because you can always check in code where class is used :)

api/models/constants.py
```diff
@@ -1,2 +1,43 @@
 DEFAULT_CHAR_FIELD_MAX_LEN = 255
 DEFAULT_CHOICE_CHAR_FIELD_MAX_LENGTH = 15
+
+# IMPORTANT: the boundaries of larger ranges must match the boundaries of the smaller ones
+
+# Match string ranges that are presented to the user to actual ranges, e.g.:
+# {"6-8": (6, 9), "9-11": (9, 12), ...}
+STUDENT_AGE_RANGE_BOUNDARIES = {
+    str_range: (int(str_range.split("-")[0]), int(str_range.split("-")[1]) + 1)
```
Roxe322, line 9: In my opinion, if you just will do it as you wrote in example, it will be more readable.

`STUDENT_AGE_RANGE_BOUNDARIES  = {"6-8": (6, 9), "9-11": (9, 12), ...}`

api/models/constants.py
```diff
@@ -1,2 +1,43 @@
 DEFAULT_CHAR_FIELD_MAX_LEN = 255
 DEFAULT_CHOICE_CHAR_FIELD_MAX_LENGTH = 15
+
+# IMPORTANT: the boundaries of larger ranges must match the boundaries of the smaller ones
+
+# Match string ranges that are presented to the user to actual ranges, e.g.:
+# {"6-8": (6, 9), "9-11": (9, 12), ...}
+STUDENT_AGE_RANGE_BOUNDARIES = {
```
Roxe322, line 8: I advise to rename it to `STUDENT_AGE_RANGES`. It will have same meaning, but shorter  => easier to read

Same below for teachers'

api/models/languages_levels.py
```diff
@@ -1,21 +1,25 @@
 from django.db import models
 
-from api.models.base import InternalModelWithName
-
 # Sometimes the languages and levels are needed separately, sometimes in combinations.
 # This means that we cannot create a "choices" field for language or level.
 
 
-class Language(InternalModelWithName):
+class Language(models.Model):
```
Roxe322, line 7: Why you decided to get rid of InternalModelWithName? 

I agree with this solution, just interesting

api/migrations/0002_data_migration.py
lemontree210: OK :)

lemontree210: I will also replace `sqlite3` with `postgresql` to be clear that that's what we'll be using in production

api/models/age_ranges.py
lemontree210: OK

api/models/languages_levels.py
lemontree210: see commit message in afb89c99a314dc3777a5ae15c3467fcbd5712b58 :)

api/models/constants.py
lemontree210: OK. My idea was that `...RANGES` would imply `range()` objects, which could be confusing if there are no `range` objects there

api/models/constants.py
lemontree210: OK. Just tried to avoid repetition

api/models/constants.py
```diff
@@ -1,2 +1,69 @@
 DEFAULT_CHAR_FIELD_MAX_LEN = 255
 DEFAULT_CHOICE_CHAR_FIELD_MAX_LENGTH = 15
+
+# IMPORTANT: the boundaries of larger ranges must match the boundaries of the smaller ones
+
+# Match string ranges that are presented to the user to actual ranges.
+# A dict comprehension would be more concise but less readable.  Dict comprehension would also be
+# less error-prone, but the dict isn't supposed to be changed frequently (or at all) after launch.
+STUDENT_AGE_RANGES = {
+    "5-6": (5, 7),
```
Roxe322, line 10: I don't understand why here and below you don't use (5, 6), (7, 8), ... etc.? Why key is 5-6, but value is 5-7? Could you explain this please? :)

lemontree210: e.g. user chooses "5-6" and the bot passes this "5-6" to the back-end. But it actually means `range (5, 7)` because 5 years old is OK, 6 years old is OK, but 7 years old is too old. `range (5, 6)` would exclude `6`.

Of course, maybe I'm thinking too much. Maybe a DB query would be just "more or equal than 5 and less or equal than 6".

But then I should go back to the list comprehension.

Roxe322: I don't think that it will be used in bare `range()`, but in orm-queries or direct comparing like `a < x < b`, so probably better to make them same. 

Also I still don't sure that you need generator here, because ranges are not same. For example, if you have identical range of 5 years, better to generate it like
`ranges = {f"{left}-{right}": (left, right) for left, right in zip(range(5, 15), range(6, 16))}`
(it will produce `{'5-6': (5, 6), '6-7': (6, 7), '7-8': (7, 8), ...}` to 15-16)

But if ranges are different, like here (from 5 to 6, but from 36 to 40), you will need to store plain ranges in code (as str-s or as tuples), and generate dict from them, so it will be more readable if you will just make dict (how it's made now)

Another option is to generate different parts of ranges (with different step), and then concatenate them.
But I still don't think that it's good idea :)

api/models/constants.py
```diff
@@ -1,2 +1,50 @@
...
+        (9, 12),
+        (13, 17),
+        (18, 95),
+    )
+}
+
+# age ranges for matching algorithm
+# TODO for now leaving these ranges the same as the ranges that students get when registering,
+#  but this is likely to change.  The children's ranges will most likely remain the same, though.
+STUDENT_AGE_RANGES_FOR_MATCHING = STUDENT_AGE_RANGES.copy()
```
Roxe322, line 50: I think we don't need to use copy here because it's constant :)

lemontree210: there will be no copy, it's a temporary thing, a stop-gap :) I can remove `.copy()` but that will be reworked anyway

Roxe322: Never mind

Roxe322 (approved):

PR #11, opened 2023-03-08
Implement views for students

natalyavalerievna:
addresses part of #30

PR #13, opened 2023-03-15
Feature/registration endpoint

Roxe322:


pyproject.toml
```diff
@@ -13,6 +13,8 @@ gunicorn = "^20.1.0"
 django-phonenumber-field = {extras = ["phonenumberslite"], version = "^7.0.2"}
 django-extensions = "^3.2.1"
 psycopg2-binary = "^2.9.5"
+djangorestframework = "^3.14.0"
+drf-yasg = "^1.21.5"
```
lemontree210, line 17: From drf_yasg docs:

> If you are looking to add Swagger/OpenAPI support to a new project you might want to take a look at [drf-spectacular](https://github.com/tfranzel/drf-spectacular), which is an actively maintained new library that shares most of the goals of this project, while working with OpenAPI 3.0 schemas.

Should we maybe use drf_spectacular then?

lemontree210 (commented):
1 nit + 1 general question :)

api/views/personal_info.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import viewsets
+
+from ..models import PersonalInfo
+from ..serializers import PersonalInfoSerializer
```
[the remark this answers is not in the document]
Roxe322: That's my bipolar. In one place I used relative import, in another - absolute. :) Will update here

pyproject.toml
Roxe322: Yeah, fixed

lemontree210 (commented):
Nit-picking only. I've had no time to run locally yet. A general question: it seems that your views are replacing the ones in #11 ? That's fine, since #11 is sort of stale now, just asking.

api/schema.py
```diff
@@ -0,0 +1,8 @@
+from drf_spectacular.openapi import AutoSchema
+
+
+class TaggedByViewSetNameSchema(AutoSchema):
+    def get_tags(self) -> [str]:
```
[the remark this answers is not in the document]
Roxe322: Just a typo :) 

In original signature this function returns list, so I agree with you, but don't think that modifying original signature is good idea :)

api/urls.py
```diff
@@ -1,10 +1,20 @@
...
+    TeacherUnder18ViewSet,
+    TeacherViewSet,
+)
 
 router = DefaultRouter()
 router.register(r"personal_info", PersonalInfoViewSet, basename="personal_info")
+router.register(r"students", StudentViewSet, basename="students")
+router.register(r"teachers", TeacherViewSet, basename="teachers")
+router.register(r"teachers_under_18", TeacherUnder18ViewSet, basename="teachers_under_18")
+router.register(r"age_ranges", AgeRangeViewSet, basename="age_ranges")
```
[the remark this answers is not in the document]
Roxe322: It's because this first param (prefix) is used later in r-string to produce regexp. You could check it in SimpleRouter source code

api/urls.py
```diff
@@ -0,0 +1,21 @@
...
+    AgeRangeViewSet,
+    PersonalInfoViewSet,
+    StudentViewSet,
+    TeacherUnder18ViewSet,
+    TeacherViewSet,
+)
+
+router = DefaultRouter()
+router.register(r"personal_info", PersonalInfoViewSet, basename="personal_info")
+router.register(r"students", StudentViewSet, basename="students")
```
lemontree210, line 14: A purely aesthetic question: it's strange why "Language(s) the students and teachers can speak in class" comes before all other fields when I browse to `/api/personal_info/` or why "Comment" comes right after "Personal info" in `/api/students/`. This doesn't matter of course, I'm just curious as to how the fields are ordered.

lemontree210:
BTW I guess we should expose the API endpoints in a README somewhere? It could be done in the same PR.

api/urls.py
Roxe322: It's order which you defined in models' fields :D 

And if you use some nested serializers, I think, they are parsed first (you could check it after latest PR update with readable nested structures instead ids), and then model's fields are parsed

Roxe322:
> BTW I guess we should expose the API endpoints in a README somewhere? It could be done in the same PR.

If I understand you correctly - you ask to add all endpoints to readme? I don't think it's necessary, because we have swagger/redoc documentation. 
I added comment with these documentation urls to readme

api/schema.py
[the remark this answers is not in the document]
lemontree210: Done in #14 . Marking this thread as resolved

api/urls.py
lemontree210: Oh, I see, `communication_language_mode` is defined in the abstract model that people and `Group` subclass from. That's why it comes first.

lemontree210:
> If I understand you correctly - you ask to add all endpoints to readme?

No of course not, I meant adding links to redoc/swagger pages - which you just did :)

README.md
```diff
@@ -1,6 +1,10 @@
 # django-webapps
 Backend for the database, implemented in Django
 
+
+## API Documentation
+Could be found at `/api/docs/swagger` or `/api/docs/redoc` after deployment (or local running).
```
lemontree210, line 6:
```suggestion
Can be found at `/api/docs/swagger` or `/api/docs/redoc` after deployment (or when running locally).
```

api/schema.py
```diff
@@ -0,0 +1,28 @@
...
+from api.serializers.errors import ValidationErrorSerializer
+
+
+class CustomSchema(AutoSchema):
+    """
+    Differences from default schema:
+    - Default schema uses same tag (from url) for all endpoints.
+      To make it more clear, we use viewset name as additional tag.
+    - We add 400 response for all endpoints with unsafe method.
+      (Also some other responses can be added here in the future, as 403 etc.).
```
lemontree210, line 13:
```suggestion
      (Also some other responses can be added here in the future, such as 403 etc.).
```

api/schema.py
```diff
@@ -0,0 +1,28 @@
...
+from rest_framework.permissions import SAFE_METHODS
+
+from api.serializers.errors import ValidationErrorSerializer
+
+
+class CustomSchema(AutoSchema):
+    """
+    Differences from default schema:
+    - Default schema uses same tag (from url) for all endpoints.
+      To make it more clear, we use viewset name as additional tag.
```
lemontree210, line 11: OK great. Maybe just a small example of how the url will be different from default.
Maybe it's clear for guys with lots of API experience, just not for a noob like me :)

api/schema.py
```diff
@@ -0,0 +1,8 @@
+from drf_spectacular.openapi import AutoSchema
+
+
+class TaggedByViewSetNameSchema(AutoSchema):
+    def get_tags(self) -> [str]:
+        tags = super().get_tags()
+        view_set_tag_name = self.view.__class__.__name__.replace("ViewSet", "")
```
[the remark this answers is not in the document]
lemontree210: Great, thanks

api/views/mixins.py
```diff
@@ -0,0 +1,24 @@
...
+
+class ReadWriteSerializersMixin:
+    """
+    A view mixin that provides a way to specify different serializers for
+    input and output purposes.
+    """
+
+    # Typings for attributes which are set in ViewSet
+    request: request.Request
+    serializer_class: type[serializers.BaseSerializer] | None
```
lemontree210, line 13: Why am I still using `Union` when pipe was already introduced in 3.10? :)

api/views/mixins.py
```diff
@@ -0,0 +1,24 @@
+from rest_framework import request, serializers
+from rest_framework.permissions import SAFE_METHODS
+
+
+class ReadWriteSerializersMixin:
```
lemontree210, line 5: That's interesting - isn't anything available out of the box in DRF?

api/views/personal_info.py
```diff
@@ -1,9 +1,25 @@
...
 
 
 class PersonalInfoViewSet(viewsets.ModelViewSet):
     queryset = PersonalInfo.objects.all()
-    serializer_class = PersonalInfoSerializer
+
+    def get_serializer_class(self) -> type[BaseSerializer]:
+        if self.action == "check_existence":
+            return PersonalInfoCheckExistenceSerializer
+        return PersonalInfoSerializer
```
lemontree210, line 16: That's interesting. so you're basically using one `ViewSet` for 2 different purposes. Well why not.

api/views/personal_info.py
```diff
@@ -1,9 +1,25 @@
...
+
+    @action(detail=False, methods=["post"])
```
lemontree210, line 18: So we're saying that the bot is going to try and create a personal info item (method `POST`) and fails if it exists? 
I was thinking that bot would send a `GET` request with name and email and if they exist, it will abort the registration. But if they do not exist, it will collect more data from the user (and only then send the whole personal info to the back end). So the existence check is not meant to be at the time when _all_ personal info is collected. I don't want to waste the (duplicate) user's time :)

Or maybe I'm misunderstanding something.

lemontree210 (commented):
mostly nitpicks and general questions that check my understanding :)

lemontree210:
Oh and I think the PR could be renamed into something a bit more informative :)

api/views/mixins.py
Roxe322: Nope. Other option is to redefine it in every viewset, or write common method to, for example, get serializer class by action (list/retrieve/etc.). And I'm not sure we need it now, because we have only two cases - read and write.

api/views/personal_info.py
Roxe322: You could read here about it - https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing

It's kinda breaking "rest", but still good for additional routes inside one "api resource"

api/views/personal_info.py
Roxe322: I think you're misunderstanding :)

We have one endpoint for creating PersonalInfo (POST /api/personal_info/) - it receives all the data once, so you need to get all data from user before sending the request.

I've made this endpoint to check unique constraint violation at the beginning of user registration process in bot. So when you have only first name, last name and email, you could make a request to check only this data, not all fields.

I use POST (even if it doesn't create anything) in security purposes. GET params are not encrypted in https, they are stored in logs, etc. etc., so better to send it as request body.

api/views/mixins.py
Roxe322: <buzzkill mode>Also it was introduced in 3.9, not 3.10</buzzkill mode>

Roxe322:
> Oh and I think the PR could be renamed into something a bit more informative :)

Fixed it a little bit, hope it's better now, I'm not sure :D

api/schema.py
Roxe322: Added

api/views/mixins.py
Roxe322: I was wrong. Really 3.9 🤦

api/views/personal_info.py
lemontree210: OK, that's good to know, thanks.

api/schema.py
```diff
@@ -0,0 +1,29 @@
+from drf_spectacular.openapi import AutoSchema
+from rest_framework.permissions import SAFE_METHODS
+
+from api.serializers.errors import ValidationErrorSerializer
+
+
+class CustomSchema(AutoSchema):
+    """
```
lemontree210, line 8: this is something I learnt recently: a multiline docstring [should contain a summary line](https://peps.python.org/pep-0257/#multi-line-docstrings) that fits on one line, followed by a blank line, followed by the more detailed explanation.

api/views/mixins.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    A view mixin that provides a way to specify different serializers for
```
lemontree210, line 7: same (summary line)

api/schema.py
Roxe322: I agree here

api/views/mixins.py
Roxe322: But don't agree here. I think this comment isn't really "multiline" - it is a summary, but because we have max line length restriction, it's split into multiple lines

lemontree210: Hehe :) I guess PEP means that this line should be one line within a limit of one line :)
So we could add one more short line on top.
But this is not really that necessary, of course.

Roxe322: Just rephrased it to make shorter and fit into one line :)

django_webapps/settings_sample.py
```diff
@@ -125,3 +130,19 @@
...
+    "SERVE_INCLUDE_SCHEMA": False,
+}
+
+#  If you import something from rest_framework before REST_FRAMEWORK setting,
+#  this setting will be ignored.
+from rest_framework import viewsets  # noqa
+
+# This is needed to be able to use annotations like "viewsets.ReadOnlyModelViewSet[Teacher]"
+# for mypy to stop ****ing our brains.
+django_stubs_ext.monkeypatch(extra_classes=[viewsets.ModelViewSet, viewsets.ReadOnlyModelViewSet])
```
lemontree210, line 148:
```suggestion

REST_FRAMEWORK = {
    "DEFAULT_SCHEMA_CLASS": "api.schema.CustomSchema",
}

SPECTACULAR_SETTINGS = {
    "SERVE_INCLUDE_SCHEMA": False,
}


# The following is needed to be able to use annotations like "viewsets.ReadOnlyModelViewSet[Teacher]"
# for mypy to stop ****ing our brains.

#  If you import something from rest_framework before REST_FRAMEWORK setting,
#  this setting will be ignored.
from rest_framework import viewsets  # noqa
django_stubs_ext.monkeypatch(extra_classes=[viewsets.ModelViewSet, viewsets.ReadOnlyModelViewSet])
```

lemontree210: I just think it should be absolutely clear that this import is only needed for mypy

tests/tests_api/test_age_range.py
```diff
@@ -0,0 +1,42 @@
...
+    [AgeRangeType.STUDENT, AgeRangeType.TEACHER, AgeRangeType.MATCHING, None],
+)
+def test_age_range_list(admin_client, age_range_type):
+    queryset = AgeRange.objects.order_by("id")
+    if age_range_type:
+        queryset = queryset.filter(type=age_range_type)
+        query_string = f"?type={age_range_type}"
+    else:
+        query_string = ""
+    response = admin_client.get(f"/api/age_ranges/{query_string}")
```
lemontree210, line 18: is trailing slash important? Some of the tests have it after `query_string`, some don't. I guess we should be consistent.

tests/tests_api/test_language_and_level.py
```diff
@@ -0,0 +1,44 @@
...
+    else:
+        query_string = ""
+
+    response = admin_client.get(f"/api/languages_and_levels/{query_string}")
+    assert response.status_code == status.HTTP_200_OK
+    assert response.json() == [
+        {
+            "id": language_and_level.id,
+            "language": {
+                "id": language_and_level.language.id,
```
lemontree210, line 23: actually this will be equal to `language` in test parametrization, not sure if you want to test this. Also, it can't be tested for language=`None`

tests/tests_api/test_language_and_level.py
```diff
@@ -0,0 +1,44 @@
...
+                "name": language_and_level.language.name,
+            },
+            "level": str(language_and_level.level),
```
lemontree210, line 26: hm... should we rely on `__str__()` method of `LanguageLevel`? I would prefer `language_and_level.level.id` here, like you do in day and time slots. Otherwise, if we change the `__str__()` on `LanguageLevel`, we'll get a sudden bug, although `id` won't change.

tests/tests_api/test_language_and_level.py
```diff
@@ -0,0 +1,44 @@
...
+    response = admin_client.get(f"/api/languages_and_levels/{language_and_level.id}/")
+
+    assert response.status_code == status.HTTP_200_OK
+    assert response.json() == {
+        "id": language_and_level.id,
+        "language": {
+            "id": language_and_level.language.id,
+            "name": language_and_level.language.name,
+        },
+        "level": str(language_and_level.level),
```
lemontree210, line 43: .

lemontree210 (changes requested):
one or two minor comments, general thoughts, and a trailing slash nitpick :)

tests/tests_api/test_age_range.py
```diff
@@ -0,0 +1,42 @@
...
+from rest_framework import status
+
+from api.models.age_ranges import AgeRange, AgeRangeType
+
+
+@pytest.mark.parametrize(
+    "age_range_type",
+    [AgeRangeType.STUDENT, AgeRangeType.TEACHER, AgeRangeType.MATCHING, None],
+)
+def test_age_range_list(admin_client, age_range_type):
```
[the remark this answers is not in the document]
Roxe322: It's a mistake. I've fixed it by doing our own fixture and adding comment into it

tests/tests_api/test_age_range.py
```diff
@@ -0,0 +1,42 @@
...
+            "age_from": age_range.age_from,
+            "age_to": age_range.age_to,
+            "type": age_range.type,
+        }
+        for age_range in queryset
+    ]
+
+
+def test_age_range_retrieve(admin_client):
+    age_range = AgeRange.objects.first()
```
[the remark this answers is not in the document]
Roxe322: We are sure that getting one object by id allows getting any other object by id, it's just general test of retrieve method

tests/tests_api/test_age_range.py
```diff
@@ -0,0 +1,42 @@
...
+    queryset = AgeRange.objects.order_by("id")
+    if age_range_type:
+        queryset = queryset.filter(type=age_range_type)
+        query_string = f"?type={age_range_type}"
+    else:
+        query_string = ""
+    response = admin_client.get(f"/api/age_ranges/{query_string}")
+
+    assert response.status_code == status.HTTP_200_OK
+    assert response.json() == [
```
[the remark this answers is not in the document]
Roxe322: I really not sure that putting this easy logic into some function will make code more readable.

How do you think it should looks like? For example:
```python
assert check_status_and_json(response, expected_status, expected_json)
```

I don't like it, I think current approach is really less "DRY", but more readable

tests/tests_api/test_day_and_time_slots.py
```diff
@@ -0,0 +1,37 @@
...
+def test_day_and_time_slot_list(admin_client):
+    queryset = DayAndTimeSlot.objects.all()
+    response = admin_client.get("/api/day_and_time_slots/")
+    assert response.status_code == status.HTTP_200_OK
+    assert response.json() == [
+        {
+            "id": day_and_time_slot.id,
+            "time_slot": {
+                "id": day_and_time_slot.time_slot.id,
+                "from_utc_hour": str(day_and_time_slot.time_slot.from_utc_hour),
```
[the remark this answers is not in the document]
Roxe322: Agree

tests/tests_api/test_age_range.py
Roxe322: Yes, It's important. For url without query string there are should be trailing slash, and with query string - no. Otherwise, tests won't pass :)

tests/tests_api/test_language_and_level.py
Roxe322: I think current solution makes it more understable. It's very clear which field is used in response, but parametrization is only about filtering.

>Also, it can't be tested for language=None

Didn't get it? Do yo mean `language_and_level` could exist without `language` (language=None)?

tests/tests_api/test_language_and_level.py
Roxe322: Agree, we shouldn't rely on str() totally :)

tests/tests_api/test_language_and_level.py
Roxe322: Also fixed

tests/tests_api/test_age_range.py
```diff
@@ -0,0 +1,42 @@
...
+        {
+            "id": age_range.id,
+            "age_from": age_range.age_from,
+            "age_to": age_range.age_to,
+            "type": age_range.type,
+        }
```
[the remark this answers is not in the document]
Roxe322: In parametrization this None means "No any filter (query params)", just clear url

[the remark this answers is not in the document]
Roxe322: So I don't think we need to do this param mandatory

tests/tests_api/test_age_range.py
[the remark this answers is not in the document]
lemontree210: Yep, never mind :)

tests/tests_api/test_age_range.py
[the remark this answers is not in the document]
lemontree210: Sure, I get it

api/serializers/personal_info.py
```diff
@@ -0,0 +1,23 @@
...
+from rest_framework import serializers
+
+from api.models import PersonalInfo
+
+
+class PersonalInfoSerializer(serializers.ModelSerializer[PersonalInfo]):
+    class Meta:
+        model = PersonalInfo
+        fields = "__all__"
+
```
lemontree210, line 12: `...WriteSerializer`?

pyproject.toml
```diff
@@ -35,6 +44,14 @@ psycopg2-binary = "^2.9.5"
 djangorestframework = "^3.14.0"
 drf-spectacular = "^0.26.0"
 django-filter = "^22.1"
+django-filter-stubs = "^0.1.2"
+djangorestframework-stubs = "^1.10.0"
+django-stubs = "^1.16.0"
```
lemontree210, line 49: are you sure these stubs belong here, not in `dev` group?

tests/tests_api/test_day_and_time_slots.py
```diff
@@ -0,0 +1,39 @@
...
+def test_day_and_time_slot_list(api_client):
+    queryset = DayAndTimeSlot.objects.all()
+    response = api_client.get("/api/day_and_time_slots/")
+    assert response.status_code == status.HTTP_200_OK
+    assert response.json() == [
+        {
+            "id": day_and_time_slot.id,
+            "time_slot": {
+                "id": day_and_time_slot.time_slot.id,
+                # Because these fields are TimeFields, they are returned as strings
```
lemontree210, line 15:
```suggestion
                # Since these fields are TimeFields, they are returned as strings
```

tests/tests_api/test_day_and_time_slots.py
```diff
@@ -0,0 +1,39 @@
...
+def test_day_and_time_slots_retrieve(api_client):
+    day_and_time_slot = DayAndTimeSlot.objects.first()
+    response = api_client.get(f"/api/day_and_time_slots/{day_and_time_slot.id}/")
+
+    assert response.status_code == status.HTTP_200_OK
+    assert response.json() == {
+        "id": day_and_time_slot.id,
+        "time_slot": {
+            "id": day_and_time_slot.time_slot.id,
+            # Because these fields are TimeFields, they are returned as strings
```
lemontree210, line 34:
```suggestion
            # Since these fields are TimeFields, they are returned as strings
```

api/serializers/errors.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class ValidationErrorSerializer(serializers.Serializer[Model]):
+    """
+    Used only in OpenAPI schema, unless we are going to redefine DRF default errors serializer.
+    """
+
+    # This means that we can have multiple errors for any field of serializer.
+    field1 = serializers.ListField(child=serializers.CharField())
+    field2 = serializers.ListField(child=serializers.CharField())
```
lemontree210, line 12: I wonder why there are exactly two fields

tests/tests_api/test_language_and_level.py
lemontree210: Never mind, I was obviously not thinking clearly in the evening :)

lemontree210 (commented):
Thanks for the changes!

api/schema.py
```diff
@@ -0,0 +1,33 @@
...
+    def get_tags(self) -> list[str]:
+        tags = super().get_tags()
+        view_set_tag_name = self.view.__class__.__name__.replace("ViewSet", "")
+        return tags + [view_set_tag_name]
+
+    def _get_response_bodies(self, direction: str = "response") -> dict[str, Any]:
+        bodies = super()._get_response_bodies(direction)
+        if self.method not in SAFE_METHODS and direction == "response":
+            bodies["400"] = self._get_response_for_code(
+                serializer=ValidationErrorSerializer,
```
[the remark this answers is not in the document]
Roxe322: It's default DRF error format, I've just put it to our serializer for schema. And it means, in general, that any unsafe method could rise ValidationError. That's almost always true.

api/serializers/teacher_under_18.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import serializers
+
+from api.models import TeacherUnder18
+
+
+class TeacherUnder18Serializer(serializers.ModelSerializer[TeacherUnder18]):
```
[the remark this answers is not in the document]
Roxe322: For now it's the same format for read an write, and we don't plan to read teachers from API (only in any other views in Dashboard)

api/serializers/teacher.py
```diff
@@ -0,0 +1,24 @@
...
+
+class TeacherWriteSerializer(serializers.ModelSerializer[Teacher]):
+    class Meta:
+        model = Teacher
+        fields = "__all__"
+
+
+class TeacherReadSerializer(serializers.ModelSerializer[Teacher]):
+    student_age_ranges = AgeRangeSerializer(read_only=True, many=True)
+    teaching_languages_and_levels = LanguageAndLevelSerializer(many=True, read_only=True)
```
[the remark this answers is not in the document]
Roxe322: Just a bipolar

api/serializers/personal_info.py
Roxe322: Not for now I think

api/serializers/errors.py
Roxe322: It could be any amount of them, but it's kind of common format to show this in schema

pyproject.toml
Roxe322: They are definitely `dev`. fixed

api/serializers/personal_info.py
lemontree210: I'm just trying to understand why you created `..WriteSerializer` in one place and not in another

api/serializers/errors.py
lemontree210: I'll have to think about this a bit more, OK for now :)

Roxe322: You could check how it looks like for "jsonfield" (which could have any fields inside) in schema by default :)

Roxe322:
> forgot to `poetry lock`? :)
Always forgetting :(

api/serializers/personal_info.py
lemontree210: So it's different for `Teacher`, but same for `TeacherUnder18`?

tests/tests_api/test_age_range.py
lemontree210: OK yes, right

tests/tests_api/test_docs.py
lemontree210, line 1: Two comments:

1. The test passes, but YAML file doesn't get built when I `runserver` and try to look at the docs
2. The test gives me a collision warning for enumerations ("status", yes, they are the same for teachers, coordinators, students). I read it [could be fixed](https://drf-spectacular.readthedocs.io/en/latest/faq.html#i-get-warnings-regarding-my-enum-or-my-enum-names-have-a-weird-suffix).

api/serializers/errors.py
lemontree210: Well, I get the `serializers.ListField(child=serializers.CharField())` bit, but I don't understand why you have `field1` and `field2`, exactly two fields. Sorry :)

Roxe322: Could you offer any other option to show that there are could be any amount of fields? For me these two "unnamed" fields show it fine :)

api/serializers/personal_info.py
Roxe322: Yes, because TeacherUnder18 doesn't have fields which should have different format on reading and writing:
`student_age_ranges, teaching_languages_and_levels, availability_slots`

api/serializers/teacher_under_18.py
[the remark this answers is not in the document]
Roxe322: Replied below for teacher

tests/tests_api/test_docs.py
Roxe322: 1. It's strange. I don't have such issue, and also we have test for docs schema building. Which error do you see?

2. I saw it, but I prefer to ignore such warnings :) I've tried to fix it, but didn't manage to. 
as doc says `If you have multiple semantically distinct enums that happen to have the same set of values, and you want different names for them, this mechanism won’t work.`
It's what we have now for Student and Teacher. I think it will be fixed after adding real statuses :)

api/serializers/errors.py
Roxe322: Example which I mean is how JsonField example is rendered by spectacular by default 
image

tests/tests_api/test_docs.py
lemontree210: 1. I just get redirected to login, as if the page didn't exist. Logger shows status 302 redirected
2. OK, got it

api/serializers/errors.py
lemontree210: But what if there are more than 2 fields? Is the developer supposed to subclass the `ValidationErrorSerializer`? Two fields `field1` and `field2` are like a magic number to me.

How does AutoSchema know what to do with these `field...` attributes of the serializer?

Damn, I don't like `drf-spectacular` docs at all. No meaningful docstrings even for classes. Makes it very hard to understand some parts. That's why maybe I completely misunderstand the issue. 

(BTW we could alias these `serializers` in import so that it's immediately clear that this is not our serializer, but that's just a thought)

api/schema.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class CustomSchema(AutoSchema):
+    """
+    Custom schema for drf-spectacular.
+
+    Differences from default schema:
+    - Default schema uses same tag - first non-parameter path part as tag for all endpoints.
+      (e.g. /api/users/ -> tag: api and also /api/other/ -> tag: api).
+      To make it more clear, we use viewset name as additional tag.
```
lemontree210, line 16:
```suggestion
      To make it more clear, we override the get_tags() method to include viewset name as additional tag.
```

api/schema.py
```diff
@@ -0,0 +1,33 @@
...
+    - We add 400 response for all endpoints with unsafe method.
```
lemontree210, line 17:
```suggestion
    - We override _get_response_bodies() method to add 400 response for all endpoints with unsafe methods.
```

Also, can and endpoint have an unsafe method? Or do you mean "for all requests to endpoints with unsafe methods"?

tests/tests_api/test_docs.py
Roxe322: Because right url is "/docs/swagger/", as in tests, without "api/" prefix. It was an error in readme, fixed it :)

api/serializers/errors.py
Roxe322: I think we should discuss it by voice, because I think you don't understand in general why I wrote this serializer  :)

api/schema.py
Roxe322: > for all requests to endpoints with unsafe methods

This is definitely more correct. Fixed

api/serializers/errors.py
lemontree210: Yep, the plan is to make me understand this, then tweak the docstring so that any potential reader understands this :)

lemontree210: I guess my question is why you decided not do use the "standard" approach mentioned in [the docs](https://www.django-rest-framework.org/api-guide/exceptions/#custom-exception-handling).

tests/conftest.py
```diff
@@ -18,3 +21,32 @@ def enable_db_access_for_all_tests(
...
+        "communication_language_mode": faker.random_element(
+            PersonalInfo.CommunicationLanguageMode.values
+        ),
+        "first_name": faker.first_name(),
+        "last_name": faker.last_name(),
+        "tg_username": faker.user_name(),
+        "email": faker.email(),
+        "phone": faker.numerify("+791########"),
+        "utc_timedelta": "03:00:00",
+        "information_source": faker.pystr(),
```
lemontree210, line 46: maybe use a [Lorem](https://faker.readthedocs.io/en/master/providers/faker.providers.lorem.html) for this?

tests/tests_api/test_personal_info.py
```diff
@@ -0,0 +1,61 @@
+"""
+We are testing only `create`, `update`, `check_existence` actions,
+because other actions are not supposed to be used.
+
+TODO:
+ Maybe we should remove some actions from ViewSet?
+ Needs to be discussed for all viewsets.
```
lemontree210, line 7: OK, maybe this can be done in a separate issue and a separate PR

lemontree210 (commented):
I will have to review once more later today when I have more time :)

lemontree210:
Another general question about Swagger. There are two things that I have doubts about: the automatic docstring shows docstring of the mixin, and then I don't get what should be passed as `/{personal_info}/`. The whole object? (no, that can't be right) ID of personal info?

image

tests/tests_api/test_personal_info.py
```diff
@@ -0,0 +1,61 @@
...
+    response = api_client.post(
+        "/api/personal_info/check_existence/",
+        data={
+            "email": existing_personal_info.email,
+            "first_name": existing_personal_info.first_name,
+            "last_name": existing_personal_info.last_name,
+        },
+    )
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+    assert response.json() == {"non_field_errors": ["User with this information already exists."]}
```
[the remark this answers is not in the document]
Roxe322: I don't think they are. Because field error is connected to some field. Not to multiple fields, as here :)

Roxe322:
> the automatic docstring shows docstring of the mixin
Fixed

 

> what should be passed as /{personal_info}/

ID of personal_info :)

tests/conftest.py
Roxe322: Seems good. Fixed to .text()

tests/tests_api/test_personal_info.py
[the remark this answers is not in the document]
Roxe322: https://github.com/samanthasgroup/django-webapps/blob/5b6fd87727069eebb2ddbc80945ce48f3cdb087d/api/serializers/personal_info.py#L15 
here it is

lemontree210:
> > what should be passed as /{personal_info}/
> 
> ID of personal_info :)

Is there a way to make def-spectacular show something like `{personal_info.id}` here? If not, maybe a quick note in README would be good.

tests/tests_api/test_student.py
```diff
@@ -0,0 +1,91 @@
...
+    # Changing for further filtering
+    for field in m2m_fields:
+        data[f"{field}__in"] = data.pop(field)
+
+    assert Student.objects.filter(**data).exists()
+
+
+def test_student_retrieve(api_client):
+    student = baker.make(Student, make_m2m=True)
+    # TODO: Student doesn't have `id` field, but `personal_info` does. Is it ok?
```
lemontree210, line 45: `Student`, `Teacher`, `Coordinator` have one-on-one relationship to `PersonalInfo` and use `PersonalInfo` as primary key. I think that's fine?

pyproject.toml
```diff
@@ -16,6 +16,21 @@ ignore_missing_imports = true
...
+disallow_untyped_calls = false
+warn_return_any = false
+
+[[tool.mypy.overrides]]
+module = [
+  "tests.*"
+]
+ignore_errors = true
+
+[mypy.plugins.django-stubs]
```
lemontree210, line 28:
```suggestion
[tool.mypy.plugins.django-stubs]
```
or not?

lemontree210 (commented):
some minor notes on the rest of the code

pyproject.toml
Roxe322: I think if it works - we shouldn't touch it :D

Roxe322: I've tested and your solution breaks it :)

api/serializers/teacher_under_18.py
```diff
@@ -3,7 +3,16 @@
 from api.models import TeacherUnder18
 
 
-class TeacherUnder18Serializer(serializers.ModelSerializer[TeacherUnder18]):
+class TeacherUnder18WriteSerializer(serializers.ModelSerializer[TeacherUnder18]):
+    class Meta:
+        model = TeacherUnder18
+        exclude = (
+            "status",
+            "comment",
```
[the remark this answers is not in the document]
Roxe322: Ok, I'll move it include it back :)

tests/tests_api/test_personal_info.py
```diff
@@ -0,0 +1,61 @@
...
+    existing_personal_info = baker.make(PersonalInfo)
+    response = api_client.post(
+        "/api/personal_info/check_existence/",
+        data={
+            "email": existing_personal_info.email,
+            "first_name": existing_personal_info.first_name,
+            "last_name": existing_personal_info.last_name,
+        },
+    )
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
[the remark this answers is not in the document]
Roxe322: This api route is needed for checking existence of personal info before the full registration.
I make existing model with `baker.make(PersonalInfo)` and then check, that for existence api returns 400

lemontree210:
Thank you for your patience 🤝 😀

tests/tests_api/test_student.py
```diff
@@ -0,0 +1,91 @@
...
+        LanguageAndLevel.objects.last().id,
+    ]
+    availability_slots_ids = [
+        DayAndTimeSlot.objects.first().id,
+        DayAndTimeSlot.objects.last().id,
+    ]
+    m2m_fields = [
+        "teaching_languages_and_levels",
+        "availability_slots",
+    ]
```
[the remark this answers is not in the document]
Roxe322: I've put it to variable to make it more clear - why we rewriting these fields filtering. Probably it could be a comment, but I think it's better with variable :)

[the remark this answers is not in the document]
Roxe322: Moved it closer to loop

tests/tests_api/test_student.py
```diff
@@ -0,0 +1,95 @@
...
+        "personal_info": personal_info.id,
+        "age_range": age_range_id,
+        "teaching_languages_and_levels": teaching_languages_and_levels_ids,
+        "is_member_of_speaking_club": faker.pybool(),
+        "requires_help_with_cv": faker.pybool(),
+        "availability_slots": availability_slots_ids,
+    }
+    response = api_client.post("/api/students/", data=data)
+
+    # TODO why does the test NOT fail when we pass no status and DOES fail when I pass one
```
Roxe322, line 29: Explained in DM. Because status is excluded from StudentWriteSerializer

lemontree210 (approved):
LGTM!

PR #14, opened 2023-03-16
Install `mypy`, add pre-commit hook, fix issues detected by `mypy` (including several methods in models)

lemontree210:
I also made sure `pyproject.toml` is compatible with #13

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,21 @@ repos:
     hooks:
     -   id: isort
         name: isort
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.1.1
+    hooks:
+      - id: mypy
+        exclude: ^(tests|api/models)/
+        args: [--strict, --strict-optional]
+        additional_dependencies:
+            - "django-stubs"
+            - "djangorestframework"
+            - "djangorestframework-stubs"
+            - "django-phonenumber-field"
+            - "phonenumberslite"
+            - "django_extensions"
+            - "drf-spectacular"
+            - "django-filter"
 -   repo: https://github.com/charliermarsh/ruff-pre-commit
     # Ruff version.
     rev: 'v0.0.254'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
             - "django-filter"
 -   repo: https://github.com/charliermarsh/ruff-pre-commit
     # Ruff version.
-    rev: 'v0.0.254'
+    rev: 'v0.0.257'
     hooks:
     -   id: ruff
         name: ruff
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,6 @@ repos:
     rev: v1.1.1
     hooks:
       - id: mypy
-        exclude: ^tests/
         args: []
         additional_dependencies:
             - "django-stubs"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
     hooks:
       - id: mypy
         exclude: ^tests/
-        args: [--strict, --python-version=3.10]
+        args: []
         additional_dependencies:
             - "django-stubs"
             - "djangorestframework"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,6 @@ repos:
     rev: v1.1.1
     hooks:
       - id: mypy
-        args: []
         additional_dependencies:
             - "django-stubs"
             - "djangorestframework"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
     rev: v1.1.1
     hooks:
       - id: mypy
-        exclude: ^(tests|api/models)/
+        exclude: ^tests/
         args: [--strict, --strict-optional, --python-version=3.10]
         additional_dependencies:
             - "django-stubs"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
     hooks:
       - id: mypy
         exclude: ^(tests|api/models)/
-        args: [--strict, --strict-optional]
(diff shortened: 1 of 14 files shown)


.pre-commit-config.yaml
```diff
@@ -9,6 +9,21 @@ repos:
     hooks:
     -   id: isort
         name: isort
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.1.1
+    hooks:
+      - id: mypy
+        exclude: ^tests/
```
lemontree210, line 16: it would be great to put a regex that would say "tests/" or "api/migrations" (or at least "api/migrations") but I can't figure out how to do it :)

pyproject.toml
lemontree210, line 1: @Roxe322 I sorted the tools alphabetically, so `black`, `isort` got moved around, but with no changes. Maybe I shouldn't have done that because that might make merging into your branch harder. But I made the whole file compatible with yours in #13.

api/models/__init__.py
```diff
@@ -18,3 +18,18 @@
     TeacherLogEvent,
 )
 from .people import Coordinator, PersonalInfo, Student, Teacher, TeacherUnder18
+
+# for mypy: listing only models that appear in admin.my
+__all__ = [
```
Roxe322, line 23: `.py` I think :)

.pre-commit-config.yaml
```diff
@@ -9,6 +9,21 @@ repos:
...
+        args: [--strict, --strict-optional, --python-version=3.10]
```
Roxe322, line 17: Why not 3.11? :)

django_webapps/settings_sample.py
```diff
@@ -10,6 +10,10 @@
 For the full list of settings and their values, see
 https://docs.djangoproject.com/en/4.1/ref/settings/
 """
+# This import is for mypy pre-commit hook. Otherwise, list[str] in ALLOWED_HOSTS triggers an error
+# despite Python version being specified in the args. See:
+# https://mypy.readthedocs.io/en/stable/runtime_troubles.html#future-annotations
```
Roxe322, line 15: Are you sure about it? I've just tested, for me it works fine without this import

.pre-commit-config.yaml
Roxe322: strict_optional is true by default:
https://mypy.readthedocs.io/en/stable/config_file.html#confval-strict_optional

.pre-commit-config.yaml
Roxe322: It should be regexp, I think something like 
`exclude: "(^tests/)|(api/migrations)|"` etc.

pyproject.toml
```diff
@@ -1,3 +1,22 @@
...
+line_length = 99
+skip_glob = ["api/migrations/*"]
+
+[tool.mypy]
+exclude = "api/migrations"
+plugins = ["mypy_django_plugin.main", "mypy_drf_plugin.main"]
+python_version = "3.10"
+strict = true
+strict_optional = true
+
```
Roxe322, line 19: I recommend to add
`ignore_missing_imports = true`
https://mypy.readthedocs.io/en/stable/config_file.html#confval-ignore_missing_imports

to avoid using type:ignore in cases such

`from phonenumber_field import modelfields  # type: ignore`

.pre-commit-config.yaml
Roxe322: Also I think migrations here are excluded by default (you could check it by removing exclude both from this file and pyproject.toml)
 
I think it's because of mypy django plugin

lemontree210: Oh, great. As for the regex, I guess `^` should be before the opening `(`, but never mind then :)

.pre-commit-config.yaml
lemontree210: Done

django_webapps/settings_sample.py
lemontree210: The funny thing is that when I run `pre-commit run -a`, it works. But when I commit in PyCharm with the hooks enabled, it fails!

pyproject.toml
lemontree210: OK

api/models/__init__.py
lemontree210: Done

django_webapps/settings_sample.py
lemontree210: You know what, I think the reason may be that when I try to commit. the pre-commit hooks run in the system-wide Python version, not in my venv (that's why you have to list all the dependencies, right?). And the system-wide Python version on Ubuntu 20.04 is 3.8. I have 3.10 installed too, obviously, but when I type `python3`, I get Python 3.8.

lemontree210: But I'm still thinking... maybe I should just do `from typing import List`

pyproject.toml
lemontree210: My reasoning was that right now this `type: ignore` thing is only needed in one place while this setting will ignore all missing imports which may not be a good thing. Like, what if stubs are available and we don't go looking for them because mypy gave no warning?

django_webapps/settings_sample.py
Roxe322: AFAIK pre-commit makes it's own venv(s), so it shouldn't break. But I'm really not sure where is trouble at your side :(

Roxe322: Maybe try to do 
`pre-commit clean`
`pre-commit uninstall`
`pre-commit install`
`pre-commit run`

.pre-commit-config.yaml
```diff
@@ -9,6 +9,20 @@ repos:
     hooks:
     -   id: isort
         name: isort
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.1.1
+    hooks:
+      - id: mypy
+        args: []
```
Roxe322, line 16: you could just remove this string :)

django_webapps/settings_sample.py
lemontree210: Okay that worked. Really weird.

.pre-commit-config.yaml
lemontree210: yep

Roxe322 (approved):

PR #26, opened 2023-03-29
Amend models in 4th iteration

### LINKED ISSUE #25: Amend models (4th iteration)
Continues work after #8 

## General
- [x] Finalize age ranges after [voting](https://t.me/c/1803788812/3430)

## Log event
- [x] add `comment`? (done in #18)
- **(not in MVP, moving to #7)** ~add something like "reason" (e.g. to mark a reason why a teacher declined a group)~

## Person (common abstract model)
- [x] add `status_since` field. It was there before but was removed. Now it's clear it's not mere duplication with `LogEvent`s: some statuses (like `ONBOARDING_STALE`) are not set along with creating `LogEvent`s, but are set just because of some time has elapsed. It's still not clear whether this `status_since` information is actually useful because it gets overwritten after status change anyway. But it could be a good shortcut, and, since statuses are not supposed to be set manually, the datetime field won't be a bother for a coordinator.

## Coordinator
- [x] Add properties for the minimum threshold (5) and maximum limit (20) of number of groups being reached.

## Teacher and Student
- [x] create model `NonTeachingHelpType`
- [x] connect `Student` and `Teacher` to it via a `ManyToManyField`

Earlier we [agreed](https://t.me/c/1803788812/3505) to create boolean fields, but then it became clear that they are likely to change, despite owners' impression. For now, we can keep boolean fields for teachers' help for other teachers, but create a model for teachers' help for students. Having a model is a great way to connect `Student` and `Teacher` since it [became clear](https://t.me/c/1803788812/3816) that the students should be asked about all of these help types, not just CV preparation.

## Student
- [x] add attribute `can_read_in_english` for `Student` (only for English, as per [poll results](https://t.me/c/1803788812/3457))

## Teacher and Young Teacher
- [x] Add `is_validated` boolean parameter to detach it from statuses.
- [x] An applicant for a teacher role must be able to choose that they don't want to teach groups at all, just speaking clubs. This means that the respective attributes (frequency, any others?) must become optional. Maybe it will break the normal form (frequency will become irrelevant if there are no groups). The number of groups can be `0` now, but `PositiveSmallInteger` allows `0` anyway.
- [x] Add status with regard to speaking club (simply "active / not active")

## EnrollmentTest
- [x] remove `levels`: if a test is available for a language, we will not be asking a student to determine their own level. This item may be solved in #40 as opposed to #26 because it's closely related to enrollment test API.

lemontree210:
Closes #25


api/models/people.py
```diff
@@ -208,43 +237,63 @@ class Teacher(TeacherCommon):
...
+    # Peer support. When a new teacher is added, they cannot have these set to True unless they
     # have prior teaching experience.  However, the `.has_prior_teaching_experience` is meant
     # to stay unchanged (it describes experience before coming to Samantha Smith's Group),
     # while it's imaginable that if a teacher teaches long enough, they will be allowed to consult
     # other teachers.  So these flags being here doesn't really break the 3rd normal form.
-    can_check_syllabus = models.BooleanField(default=False)
-    can_consult_other_teachers = models.BooleanField(default=False)
-    can_give_feedback = models.BooleanField(default=False)
-    can_help_with_children_group = models.BooleanField(
+    peer_support_can_check_syllabus = models.BooleanField(
```
Roxe322, line 263: Why did you decide to add `peer_support` prefix to these fields? In my opinion, it looked better without it :)

api/models/people.py
```diff
@@ -131,6 +147,20 @@ def __str__(self) -> str:
...
+    def has_enough_groups(self) -> bool:
+        """Returns `True` if coordinator has required minimum amount of groups."""
+        if self.group_set.count() >= CoordinatorGroupLimit.MIN:
+            return True
+        return False
+
+    @property
+    def has_reached_group_limit(self) -> bool:
+        """Returns `True` if coordinator has reached maximum amount of groups."""
+        if self.group_set.count() >= CoordinatorGroupLimit.MAX:
```
Roxe322, line 160: Let's add related_name for this to be `groups`

api/models/people.py
```diff
@@ -131,6 +147,20 @@ def __str__(self) -> str:
         role = " (admin)" if self.is_admin else ""
         return f"{super().__str__()}{role}"
 
+    @property
+    def has_enough_groups(self) -> bool:
+        """Returns `True` if coordinator has required minimum amount of groups."""
+        if self.group_set.count() >= CoordinatorGroupLimit.MIN:
```
Roxe322, line 153: Here and below: `self.group_set.count() >= CoordinatorGroupLimit.MIN` is already a bool so you can just do 
`return self.group_set.count() >= CoordinatorGroupLimit.MIN`

tests/tests_api/test_teacher_under_18.py
```diff
@@ -30,7 +32,9 @@ def test_teacher_under_18_retrieve(api_client):
     assert response.json() == {
         "personal_info": teacher_under_18.personal_info.id,
         "comment": teacher_under_18.comment,
-        "additional_skills_comment": teacher_under_18.additional_skills_comment,
-        "can_help_with_speaking_club": teacher_under_18.can_help_with_speaking_club,
+        "can_host_speaking_club": teacher_under_18.can_host_speaking_club,
         "status": teacher_under_18.status,
+        "status_since": teacher_under_18.status_since.strftime("%Y-%m-%dT%H:%M:%S.%fZ"),
```
Roxe322, line 37: I think some of that should work, either `str(teacher_under_18.status_since)` or `teacher_under_18.status_since.isoformat()`

tests/tests_api/test_teacher_under_18.py
```diff
@@ -11,9 +11,11 @@ def test_teacher_under_18_create(api_client, faker):
 
     data = {
         "personal_info": personal_info.id,
-        "additional_skills_comment": faker.pystr(max_chars=DEFAULT_CHOICE_CHAR_FIELD_MAX_LENGTH),
-        "can_help_with_speaking_club": faker.pybool(),
+        "can_host_speaking_club": faker.pybool(),
         "comment": faker.text(),
+        "status_since": faker.date_time(tzinfo=tz.gettz("UTC")),
```
Roxe322, line 16: I think more standard approach is to use pytz lib, because it gives us some constants, which is better than strings

api/models/people.py
Roxe322: I mean instead of group_set

api/migrations/0002_data_migration.py
```diff
@@ -108,6 +109,26 @@ def _write_languages_and_levels(self):
...
+            ("mock_interview", "Mock interview"),
+            ("job_search", "Job search"),
+            ("career_strategy", "Career strategy"),
+            ("linkedin", "LinkedIn profile"),
+            ("career_switch", "Career switch"),
+            ("portfolio", "Portfolio for creative industries"),
+            ("uni_abroad", "Entering a university abroad"),
+            ("translate_docs", "Translation of documents"),
+        ):
+            help_types.append(HelpType(id=id_and_name[0], name=id_and_name[1]))
```
Roxe322, line 128: It's better to use unpacking:

`for id, name in (...):`

api/models/non_teaching_help.py
```diff
@@ -0,0 +1,17 @@
...
+
+    Both Student and Teacher can have connections to this model, meaning that a teacher can
+    provide this kind of help, while a student requires this kind of help.
+    """
+
+    id = models.CharField(max_length=20, primary_key=True)  # for easier connection with bot
+    name = models.CharField(max_length=DEFAULT_CHAR_FIELD_MAX_LEN, unique=True)
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
Roxe322, line 17: self.name is already a string, so you don't need to put it into f-string

api/models/people.py
lemontree210: I know, I too hate prefixes that are repeated again and again. But I want these boolean fields to be clearly separated from e.g. `can_host_speaking_club`. 

These boolean fields would be potential candidates for moving to a separate model, but it doesn't seem to be justified just yet (as opposed to `NonTeachingHelp`). We're leaving them in `Teacher`, but I want them to be separated from the rest.

Roxe322: Ok, got it, let's keep it then

api/models/people.py
lemontree210: Done

api/models/people.py
lemontree210: of course 🤦🏻

tests/tests_api/test_teacher_under_18.py
lemontree210: Nah...

First solution gives:
```
{'status_since': '2023-04-09T11:54:51.645249Z'} != {'status_since': '2023-04-09 11:54:51.645249+00:00'}
```

Second solution gives:
```
{'status_since': '2023-04-09T11:53:34.353149Z'} != {'status_since': '2023-04-09T11:53:34.353149+00:00'}
```

tests/tests_api/test_teacher_under_18.py
lemontree210: Done

api/migrations/0002_data_migration.py
lemontree210: Done

api/models/non_teaching_help.py
lemontree210: Yep. I had an ID in there, and then removed it, forgot to remove f-string

api/models/people.py
lemontree210: I couldn't just put "groups" though, had to use `%('class')ss`

tests/tests_api/test_teacher_under_18.py
Roxe322: That's weird behaviour in either python's datetime or drf.

I recommend to keep it as 
teacher_under_18.status_since.isoformat().replace("+00:00", "Z")

because in DRF datetime field source code there is:
```python
        if output_format.lower() == ISO_8601:
            value = value.isoformat()
            if value.endswith('+00:00'):
                value = value[:-6] + 'Z'
            return value
```
🤯

Roxe322: I think it's because of mismatch of formats in JS and Python

Roxe322 (approved):

PR #46, opened 2023-04-17
Data migration with fake data

Roxe322:


api/models/helpers.py
lemontree210, line 1: I would probably choose a different name because `DataMigrationMaster` doesn't do any "help" for models. While the module contains only one class anyway, it's logical to just call it `data_migration_master.py` (or `data_migration_maker` if you agree with renaming of the class).

api/models/helpers.py
```diff
@@ -0,0 +1,29 @@
...
+    See Django docs: https://docs.djangoproject.com/en/stable/topics/migrations/#data-migrations
+    """
+
+    def __init__(self, apps: StateApps, schema_editor: DatabaseSchemaEditor):
+        self.apps = apps
+        self.schema_editor = schema_editor
+
+    @abc.abstractmethod
+    def main(self) -> None:
+        """Here should be placed data migration operations."""
```
lemontree210, line 19:
```suggestion
        """Override this method and put data migration operations in it."""
```

api/models/helpers.py
```diff
@@ -0,0 +1,29 @@
+import abc
+
+from django.db.backends.sqlite3.schema import DatabaseSchemaEditor
+from django.db.migrations.state import StateApps
+
+
+class DataMigrationMaster(abc.ABC):
+    """
+    Abstract class-based variation used to make data-migrations.
```
lemontree210, line 9:
```suggestion
    Abstract class-based variation used to make data migrations.
```

api/models/helpers.py
```diff
@@ -0,0 +1,29 @@
+import abc
+
+from django.db.backends.sqlite3.schema import DatabaseSchemaEditor
+from django.db.migrations.state import StateApps
+
+
+class DataMigrationMaster(abc.ABC):
```
lemontree210, line 7:
```suggestion
class DataMigrationMaker(abc.ABC):
```
We can make the name more predictable, consistent with `makemigrations`.

api/models/helpers.py
```diff
@@ -0,0 +1,29 @@
...
+    Abstract class-based variation used to make data-migrations.
+    See Django docs: https://docs.djangoproject.com/en/stable/topics/migrations/#data-migrations
+    """
+
+    def __init__(self, apps: StateApps, schema_editor: DatabaseSchemaEditor):
+        self.apps = apps
+        self.schema_editor = schema_editor
+
+    @abc.abstractmethod
+    def main(self) -> None:
```
lemontree210, line 18:
```suggestion
    def _populate(self) -> None:
```
Both `main()` and `run()` in the same class are a bit confusing. I'd suggest making this method "private" to make it absolutely clear which one should be passed to `RunPython`.

api/models/helpers.py
```diff
@@ -0,0 +1,29 @@
...
+        """Here should be placed data migration operations."""
+        pass
+
+    @classmethod
+    def run(cls, apps: StateApps, schema_editor: DatabaseSchemaEditor) -> None:
+        """
+        Runs data migration operations.
+        This method is called by Django and should be passed into RunPython.
```
lemontree210, line 26:
```suggestion
        Creates an instance with necessary methods and attributes, then runs data migration operations.
        Pass this method to `RunPython` and it will called by Django when running `makemigrations`.
        
        See Django docs: 
        https://docs.djangoproject.com/en/stable/ref/migration-operations/#runpython
```

api/migrations/0002_data_migration.py
```diff
@@ -1,31 +1,19 @@
...
-        self.apps = apps
-        self.schema_editor = schema_editor
-        # Adding __call__ to a class won't work because RunPython can't create an instance and then
-        # run it.  So we're running main() right after instantiating, which effectively makes
-        # PrePopulationMaster into a callable with instance attributes.
-        self.main()
+class InitialDataPopulationMaster(DataMigrationMaster):
 
     def main(self):
         """Runs pre-population operations."""
```
lemontree210, line 19:
```suggestion
    def _populate(self):
        """Populates the database with initial data."""
```

api/migrations/0002_data_migration.py
```diff
@@ -1,31 +1,19 @@
...
-    """
-
-    def __init__(self, apps: StateApps, schema_editor: DatabaseSchemaEditor):
-        self.apps = apps
-        self.schema_editor = schema_editor
-        # Adding __call__ to a class won't work because RunPython can't create an instance and then
-        # run it.  So we're running main() right after instantiating, which effectively makes
-        # PrePopulationMaster into a callable with instance attributes.
-        self.main()
+class InitialDataPopulationMaster(DataMigrationMaster):
```
lemontree210, line 16:
```suggestion
class InitialDataPopulator(DataMigrationMaker):
```
see comment in`helpers.py`

api/migrations/0003_fake_data.py
```diff
@@ -0,0 +1,189 @@
...
+
+APP_NAME = "api"
+
+AMOUNT_OF_COORDINATORS = 10
+AMOUNT_OF_STUDENTS = 100
+AMOUNT_OF_TEACHERS = 25
+AMOUNT_OF_TEACHERS_UNDER_18 = 5
+
+
+class FakeDataPopulationMaster(DataMigrationMaster):
```
lemontree210, line 27:
```suggestion
class FakeDataPopulator(DataMigrationMaker):
```

pyproject.toml
```diff
@@ -18,6 +18,7 @@ python_version = "3.10"
 strict = true
 disallow_untyped_calls = false
 warn_return_any = false
+disallow_subclassing_any = false
```
lemontree210, line 21: is it because of `abc.ABC`?

api/migrations/0003_fake_data.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def _make_fake_teachers(self):
+        """Makes fake teachers."""
+        self.teacher_recipe.make(_quantity=AMOUNT_OF_TEACHERS)
+
+    def _make_fake_teachers_under_18(self):
+        """Makes fake teachers under 18."""
+        self.teacher_under_18_recipe.make(_quantity=AMOUNT_OF_TEACHERS_UNDER_18)
+
+    def main(self):
```
lemontree210, line 171:
```suggestion
    def _populate(self):
```

lemontree210 (changes requested):
mostly naming nitpicks. My general idea is that it should be very clear that `.run()` is the only public method. It's totally OK if the user has to override a private method, just like you do for an m2m baker.

I don't understand the trick with `CallableM2MBaker` BTW, will ask in a PM.

pyproject.toml
Roxe322: It's because of subclassing Baker class for our CustomBaker

api/models/helpers.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    def __init__(self, apps: StateApps, schema_editor: DatabaseSchemaEditor):
+        self.apps = apps
+        self.schema_editor = schema_editor
+
+    @abc.abstractmethod
+    def main(self) -> None:
+        """Here should be placed data migration operations."""
+        pass
```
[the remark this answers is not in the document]
Roxe322: Fixed

api/models/helpers.py
Roxe322: Fixed

api/migrations/0003_fake_data.py
Roxe322: fixed

api/models/helpers.py
Roxe322: Agree, renamed it into data_migration_maker

api/migrations/0002_data_migration.py
Roxe322: Here I wanted to be consistent with DataMigrationMaker base class name, so in my mind it should be called InitialDataMaker and FakeDataMaker.

But renamed it as you suggested :)

api/migrations/0003_fake_data.py
Roxe322: Fixed

api/migrations/0002_data_migration.py
lemontree210: You know what, you're right. I think it was my mistake to call it `DataMigrationMaker` because `makemigrations` actually **creates** migration files (whereas we create the data migration module by hand), and then the migrations are applied when `migrate` is run.

On the other hand, I didn't like the `...Master` thingy because it sounded too vague. So how about calling the abstract base class `Populator` or something like that?

api/migrations/0003_fake_data.py
```diff
@@ -50,7 +66,27 @@ def _get_random_amount_of_objects(
...
+            ),
+            students=lambda: related(
+                *[self.student_recipe] * self.faker.pyint(min_value=5, max_value=25)
+            ),
+            teachers=lambda: related(
+                *[self.teacher_recipe]
+                * self.faker.pyint(
+                    min_value=teachers_min_value, max_value=teachers_max_value
+                )
+            ),
```
lemontree210, line 85: a group usually has one teacher, in rare cases two teachers. Same for coordinators.
number of students is 2 to 10 people, as per the owners' poll. I think it makes sense to follow these guidelines for the dashboard to look real right from the start.

api/migrations/0003_fake_data.py
```diff
@@ -152,29 +190,86 @@ def _make_teacher_under_18_recipe(self):
             is_validated=self.faker.pybool,
         )
 
-    def _make_fake_coordinators(self):
-        """Makes fake coordinators."""
-        self.coordinator_recipe.make(_quantity=AMOUNT_OF_COORDINATORS)
+    def __get_random_time_or_none(self) -> time | None:
```
lemontree210, line 193: double underscore? you needed name mangling here?

api/migrations/0003_fake_data.py
```diff
@@ -152,29 +190,86 @@ def _make_teacher_under_18_recipe(self):
...
+            ),
+            is_for_staff_only=self.faker.pybool,
+            language_and_level=lambda: self.faker.random_element(
+                LanguageAndLevel.objects.all()
+            ),
+            lesson_duration_in_minutes=lambda: self.faker.pyint(
+                min_value=30, max_value=120, step=30
+            ),
+            status=lambda: self.faker.random_element(GroupStatus.values),
+            start_date=self.faker.past_date,
```
lemontree210, line 212: groups can also start in the future but I agree that this is irrelevant in a context of just displaying the data in the dashboard

api/migrations/0003_fake_data.py
```diff
@@ -152,29 +190,86 @@ def _make_teacher_under_18_recipe(self):
...
+            # FIXME: Sometimes all of the days are generated as None.
+            #  It violates the model's constraint.
+            #  Just restart the migration and pray to the God of Random.
+            monday=self.__get_random_time_or_none,
+            tuesday=self.__get_random_time_or_none,
+            wednesday=self.__get_random_time_or_none,
+            thursday=self.__get_random_time_or_none,
+            friday=self.__get_random_time_or_none,
+            saturday=self.__get_random_time_or_none,
+            sunday=self.__get_random_time_or_none,
```
lemontree210, line 223: actually no group can have classes more often than 3 times a week, so if a group has 7 times all filled in, it will be an unrealistic group that might look unrealistic on the dashboard.

Maybe generate a dictionary with 2-3 days randomly chosen as keys and then do `extend` with `(**dict)`?

api/migrations/0003_fake_data.py
```diff
@@ -152,29 +190,86 @@ def _make_teacher_under_18_recipe(self):
...
+            language=lambda: self.faker.random_element(Language.objects.all()),
+            teachers_under_18=lambda: related(
+                *[self.teacher_under_18_recipe]
+                * self.faker.pyint(min_value=0, max_value=2)
+            ),
+        )
+
+    def _make_fake_coordinators_without_group(self):
+        """Makes fake coordinators without group."""
+        self.coordinator_recipe.make(_quantity=AMOUNT_OF_COORDINATORS_WITHOUT_GROUP)
```
lemontree210, line 241: looks a bit inconsequential that you're using constants here but plain numbers in `min_value` and `max_value`

api/migrations/0003_fake_data.py
```diff
@@ -0,0 +1,284 @@
...
+        )
+
+    def _make_group_common_recipe(
+        self, model_name: str, teachers_min_value: int, teachers_max_value: int
+    ) -> Recipe:
+        return Recipe(
+            model_name,
+            coordinators=lambda: related(
+                *[self.coordinator_recipe] * self.faker.pyint(min_value=1, max_value=5)
+            ),
```
lemontree210, line 76: could you explain in a short comment in code what this `related` with the unpacking does here?

lemontree210 (commented):
I finally understood the logic, but at first I got confused with the names :)

`_make_fake_groups()` calls method `.make` of `self.group_recipe`, but to produce `self.group_recipe` you have to call `self._make_group_recipe()`. 

I mean, I get it after reading it 3 times :) I'm not saying it's bad, it was just hard to understand this collection of `make`s and `recipe`s at first. Maybe some of it could be extracted into some sort of `RecipeMaker` class.

I don't insist on any of this, though. Don't want to be too nit-picking, especially when generating fake data! :)

api/migrations/0002_data_migration.py
Roxe322: renamed it to Populator.

hardest thing in programming is naming :D

api/migrations/0003_fake_data.py
Roxe322: Makes sense. Fixed

api/migrations/0003_fake_data.py
Roxe322: Just wanted to show that it's not protected but private method. Fixed it to one underscore

api/migrations/0003_fake_data.py
Roxe322: Tried to make it more simple :)

api/migrations/0003_fake_data.py
Roxe322: Fixed it in places that could be changed

api/migrations/0003_fake_data.py
Roxe322: Problem with generating the dictionary of fixed days is that then we will have same days for all groups, it's bad even if we have different time for them. I didn't manage to find any option to create this dict randomly because we can use lambda only as a field value, not as a field itself. I have an idea of doing check after recipe "making", but it definitely will look more ugly than current solution. 

So I decided to keep current approach - generate for all days but with `random_time_or_none`, it usually means that groups will have 3-4 active days and 3-4 `None` days. 

Also made a fix to be sure that needed amount of groups is created, check it out

Roxe322: We could also try to use weighting (https://faker.readthedocs.io/en/master/providers/baseprovider.html#faker.providers.BaseProvider.random_elements)
 here for choosing between None or time, but I'm not sure if it's needed

api/migrations/0003_fake_data.py
Roxe322: I've tried :)

Roxe322: Reply to the above comment: I've split logic into two classes, made an RecipeStorage class which stores recipes

api/migrations/0003_fake_data.py
```diff
@@ -37,18 +37,38 @@
...
+
+MIN_AMOUNT_OF_TEACHERS_IN_GROUP = 1
+MAX_AMOUNT_OF_TEACHERS_IN_GROUP = 2
+
+MIN_AMOUNT_OF_TEACHERS_UNDER_18_IN_SPEAKING_CLUB = 1
+MAX_AMOUNT_OF_TEACHERS_UNDER_18_IN_SPEAKING_CLUB = 2
+
+
+class RecipeStorage:
+    def __init__(self):
```
lemontree210, line 54:
```suggestion
    """Helper class for producing recipes with fake data.
    
    For the sake of readability, this class is separated from class 
    that populates database data produced from these recipes.
    """

    def __init__(self):
```

api/migrations/0003_fake_data.py
```diff
@@ -211,56 +237,75 @@ def _make_group_recipe(self) -> Recipe:
...
+        We need to recreate such objects to skip db constraints and be sure that needed amount is generated.
+        """
+        for _ in range(amount):
+            while True:
+                try:
+                    recipe.make()
+                    break
+                except DatabaseError:
+                    pass
+
```
lemontree210, line 279:
```suggestion
            while True:
                try:
                    recipe.make()
                except DatabaseError:
                    pass
                else:
                    break

```

lemontree210: No difference, of course. Obviously, there can be no error when executing `break`. I'm probably being a maximalist here (only one statement in `try`).

api/migrations/0003_fake_data.py
```diff
@@ -211,56 +237,75 @@ def _make_group_recipe(self) -> Recipe:
...
     def _make_fake_coordinators_without_group(self):
         """Makes fake coordinators without group."""
-        self.coordinator_recipe.make(_quantity=AMOUNT_OF_COORDINATORS_WITHOUT_GROUP)
+        self.recipes.coordinator.make(_quantity=AMOUNT_OF_COORDINATORS_WITHOUT_GROUP)
```
lemontree210, line 282: that looks much nicer to me!

api/models/data_populator.py
```diff
@@ -0,0 +1,28 @@
+import abc
+
+from django.db.backends.sqlite3.schema import DatabaseSchemaEditor
+from django.db.migrations.state import StateApps
+
+
+class DataPopulator(abc.ABC):
+    """
+    Abstract class-based variation used to make data migrations.
```
lemontree210, line 9:
```suggestion
    Abstract class-based variation used to apply data migrations.
```
(to discern `makemigrations` from `migrate`. This populator thing is for the `migrate` stage.

api/migrations/0003_fake_data.py
```diff
@@ -0,0 +1,329 @@
...
+        )
+
+    def _make_fake_speaking_clubs(self):
+        """Makes fake speaking_clubs."""
+        self._make_amount_of_recipe_and_skip_database_error(
+            self.recipes.speaking_club, AMOUNT_OF_SPEAKING_CLUBS
+        )
+
+    def _populate(self):
+        """Runs pre-population operations."""
```
lemontree210, line 311:
```suggestion
        """Runs operations required for populating the database with fake data."""
```

Can look a bit strange that `_populate` does "pre-population".

lemontree210 (changes requested):
The changes look great, thank you! Maybe we could even merge this PR without trying to generate **all** fake data for the dashboard right away.

api/migrations/0003_fake_data.py
```diff
@@ -0,0 +1,329 @@
...
+        self.speaking_club = self._make_speaking_club_recipe()
+
+    def _get_random_time_or_none(self) -> time | None:
+        return self.faker.random_element([None, self.faker.time_object()])
+
+    def _get_group_days_of_week(self) -> dict[str, time | None]:
+        return {
+            day_name.lower(): lambda: self._get_random_time_or_none()
+            for day_name in DayAndTimeSlot.DayOfWeek.labels
+        }
```
lemontree210, line 71: Maybe you could generate it here while there's at least 1 spot that is not `None`, and then you return it?

I wouldn't really be bothered by having **no more** than 3 slots here, as soon as we avoid the situation when all are `None`. But of course we could include the check here as well.

```
while True:
    dict_ = {
                day_name.lower(): lambda: self._get_random_time_or_none()
                for day_name in DayAndTimeSlot.DayOfWeek.labels
            }

if len(v for v in dict_.values() if v) in range(1, 4):
    return dict_
```

lemontree210 (commented):
just an idea for generating a dictionary with an acceptable number of day and time slots

api/migrations/0003_fake_data.py
Roxe322: I think with `else` it will be less readable. Just imho :)

api/migrations/0003_fake_data.py
Roxe322: It won't work as you expect. I'll try to explain, but ask me in DM if my explanation is bad :) 

We have callable (lambda) for creating some value on recipe "baking" (creating model). This lambda is called for every model instance creation.

But this method `_get_group_days_of_week` is called only once. So for your approach it will generate constant 1-4 days, which values then will be `time or None` on every model creation. 
But if it's what you excepted - than we could do it easier: 

```
{
                day_name.lower(): lambda: self._get_random_time_or_none()
                for day_name in self.faker.random_elements(DayAndTimeSlot.DayOfWeek.labels, unique=True, length=4)
            }
```


Now we have `while True` with `except DatabaseError` on recipe making, so we could be sure that at least one day is presented. And the advantage of my solution that we have random days for every group, not only selected from 4 of them.

Also, as I said earlier, we could try to use weights for `None or time` to make `None `more common and `time` more `rare`. But I'm not really sure that we need it

lemontree210: OK, I see, basically it's a difference between passing a callback and its result

lemontree210 (approved):
LGTM

Issue #30, opened 2023-03-30
Create read-only API views for ToolJet dashboard (lists and profiles)


lemontree210:
The tasks can be done step by step in small PRs. Please mention the PRs in the bullet points as you go.

## Preliminaries and Implementation details
- [x] create a separate data migration with a significant amount of dummy students, teachers, coordinators (using `faker` to create "normal" names) (closed by #46)
- ~implement token authentication (may not be needed if all requests come from within the same server, but we want to be able e.g. to test the bot locally, so that may still be needed)~ leaving for later

## General overview
- [x] all students without personal info (closed by #57)
- [x] all groups (closed by #65)
- [x] all teachers without personal info (closed by #81)
- ~all coordinators with personal info~ - not needed for dashboard, [confirmed](https://t.me/c/1803788812/4325) by owners

<details>

<summary>Example output for "all students" view (without personal info) </summary>

```json
[
    {
        "id": 1,
        "first_name": "Sergey",
        "last_name": "Borisov",
        "utc_code": "UTC+3",
        "availability_slots": [
            {
                "day_of_week": 0,
                "time_from": "18",
                "time_to": "19"
            },
            {
                "day_of_week": 2,
                "time_from": "18",
                "time_to": "19"
            }
        ],
        "communication_mode": "RU/UA",
        "age_range": "18-20",
        "speaking_club_status": "NOT_GOING",
        "status": "GROUP_ASSIGNED",
        "languages_and_levels": [
            {
                "language": "English",
                "level": "A1"
            }
        ]
    }
]
```

</details>

<details>

<summary>Example output for "all teachers" view (without personal info)</summary>

```json
[
    {
        "id": 1,
        "first_name": "Sergey",
        "last_name": "Vasilyev",
        "status": "TEACHING_ACCEPTING_MORE",
        "communication_mode": "RU/EN",
        "teacher_info": {
            "utc_code": "UTC+3",
            "availability_slots": [
                {
                    "day_of_week": 0,
                    "time_from": "18",
                    "time_to": "19"
                },
                {
                    "day_of_week": 2,
                    "time_from": "18",
                    "time_to": "19"
                }
            ],
            "student_age_ranges": [
                "18-25",
                "25-34"
            ],
            "languages_and_levels": [
                {
                    "language": "German",
                    "level": "A2"
                },
                {
                    "language": "English",
                    "level": "A1"
                }
            ],
            "max_number_of_groups": 2,
            "max_weekly_sessions": 4,
[...]

PR #57, opened 2023-04-29
Public Students API

Roxe322:
Addresses part of #30 . This is a "public" view of students for the dashboard (information about the students that any logged in coordinator can see).


api/serializers/day_and_time_slot.py
```diff
@@ -15,3 +15,12 @@ class DayAndTimeSlotSerializer(serializers.ModelSerializer[DayAndTimeSlot]):
     class Meta:
         model = DayAndTimeSlot
         fields = "__all__"
+
+
+class MinifiedDayAndTimeSlotSerializer(serializers.ModelSerializer[DayAndTimeSlot]):
```
lemontree210, line 20: add a short docstring explaining what kind of minification is done here?

Same for language and level

api/serializers/student.py
```diff
@@ -25,3 +29,32 @@ class StudentReadSerializer(serializers.ModelSerializer[Student]):
     class Meta:
         model = Student
         exclude = ("children",)
+
+
+class PublicStudentSerializer(serializers.ModelSerializer[Student]):
```
lemontree210, line 34: add a short docstring explaining what "public" is in this context
(None of the information is actually public, so I guess you mean "available to all coordinators" or "to all users of the dashboard").

api/urls.py
```diff
@@ -9,25 +9,37 @@
...
+internal_router.register(r"enrollment_test", EnrollmentTestViewSet, basename="enrollment_test")
+internal_router.register(
     r"enrollment_test_result", EnrollmentTestResultViewSet, basename="enrollment_test_result"
 )
-router.register(r"non_teaching_help", NonTeachingHelpViewSet, basename="non_teaching_help")
+internal_router.register(
+    r"non_teaching_help", NonTeachingHelpViewSet, basename="non_teaching_help"
+)
+
+public_router = DefaultRouter()
```
lemontree210, line 38: add a short comment explaining what "public" is? (Like with the class above)
Just so that someone reading this code doesn't think we actually have some *publicly* available data

lemontree210 (changes requested):
Mostly nitpicks, but also a general question about the term "public".

If by "public" you mean *any data* requested from ToolJet (and you don't return email address or phone number in those views/serializers), then it's not entirely correct because a coordinator should see personal data for the students in *their group* (while they should see no personal data for all other students), and that data will also be requested by ToolJet in a different table. My guess (by [this line](https://github.com/samanthasgroup/django-webapps/blob/feature/public_students_view/api/views/student.py#L19) in student.py) is that this is what you mean by "public".

If we use the word "public" to describe data about all students requested by any coordinator (not only about this coordinator's students), then maybe we should change the wording because none of the data we return is actually public (i.e. visible to anyone). We could use something like "sanitized" instead of "public" in this case.

If it isn't clear what I'm talking about, let's discuss it in PMs :)

api/serializers/student.py
Roxe322: I mean "internal" api is for bot and "public" is for tooljet :)

api/serializers/day_and_time_slot.py
Roxe322: Isn't it clear enough from fields presented inside serializer?

api/serializers/utc_timedelta.py
```diff
@@ -0,0 +1,13 @@
+import datetime
+
+from rest_framework import serializers
+
+
+class UTCTimedeltaField(serializers.Field):  # type: ignore
+    """Read only field that returns a string representation of a UTC timedelta."""
```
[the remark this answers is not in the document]
Roxe322: Yes, fixed

api/migrations/0003_fake_data.py
```diff
@@ -177,7 +178,7 @@ def _make_student_recipe(self) -> Recipe:
             ),
             smalltalk_test_result=self.faker.json,
             teaching_languages_and_levels=lambda: self._get_random_amount_of_objects(
-                LanguageAndLevel
+                LanguageAndLevel, max_length=4
```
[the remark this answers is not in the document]
Roxe322: Ok

api/serializers/day_and_time_slot.py
Roxe322: Nevertheless, I've added comments

api/serializers/student.py
Roxe322: But added comment

api/urls.py
Roxe322: Done

api/serializers/day_and_time_slot.py
lemontree210: Thanks! Well I was more thinking about IDE hints :)

lemontree210 (approved):
Thanks for the changes!

I meant that Tooljet will also have a "private" view where a coordinator will have all the information about the students and teachers in their group.

But we can leave the renaming for later, I don't want to slow down the development because of that.

PR #65, opened 2023-05-02
Public Group View

Roxe322:
addresses part of #30


api/serializers/coordinator.py
```diff
@@ -0,0 +1,14 @@
+from rest_framework import serializers
+
+from api.models import Coordinator
+
+
+class MinifiedCoordinatorSerializer(serializers.ModelSerializer[Coordinator]):
+    """Serializer for Coordinator model with only id and full_name fields."""
+
+    full_name = serializers.CharField(source="personal_info.full_name")
+    id = serializers.IntegerField(source="personal_info_id")
```
lemontree210, line 10: `_id`, not `.id`? Same for teacher

api/serializers/group.py
```diff
@@ -0,0 +1,39 @@
...
+            "sunday",
+            "language_and_level",
+            "lesson_duration_in_minutes",
+            "status",
+            "start_date",
+            "end_date",
+            "telegram_chat_url",
+            "coordinators",
+            "teachers",
+            "students_count",  # This field should come from annotation.
```
lemontree210, line 37:
```suggestion
            "students_count",  # This field will be added in ViewSet by annotating a QuerySet.
```
BTW that is a sort of "circular dependency" (serializer knows about view functionality, view imports serializer), but maybe there's no other way.

api/serializers/coordinator.py
Roxe322: it's same :)

api/views/group.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
+from rest_framework import viewsets
+
+from api.models import Group
+from api.serializers import PublicGroupSerializer
+
+
+class PublicGroupViewSet(viewsets.ReadOnlyModelViewSet[Group]):
```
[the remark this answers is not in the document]
Roxe322: Let's do it step by step. For now it's read-only, and it could be changed in future when we will need it :)

api/serializers/group.py
Roxe322: Another option is to use "SerializerMethodField" (or property in model) to count students for each group. It's much slower and inefficient so current solution is better.

api/serializers/coordinator.py
Roxe322: I mean, `personal_info_id` it's still a field of model, which stores id of related personal_info. 
Usually, when you need only id, it's better to use it as I wrote it, because when you write `related_model.id`, django should hit database, load `related_model` and only then get a field. 
It's not a problem here, because we still use other fields of `personal_info`, just general comment.

lemontree210 (approved):

PR #66, opened 2023-05-02
populate enrollment test

### LINKED ISSUE #62: Create data migration with 2 English assessments

lemontree210:
closes #62

### DIFF
--- a/api/migrations/0002_data_migration.py
+++ b/api/migrations/0002_data_migration.py
@@ -161,8 +161,6 @@ class InitialDataPopulator(DataPopulator):
                     is_correct=False,
                 )
 
-            enrollment_test.save()
-
     def _write_non_teaching_help(self):
         HelpType = self.apps.get_model(APP_NAME, "NonTeachingHelp")
 
--- a/api/migrations/0002_data_migration.py
+++ b/api/migrations/0002_data_migration.py
@@ -110,18 +110,9 @@ class InitialDataPopulator(DataPopulator):
         AgeRange = self.apps.get_model(APP_NAME, "AgeRange")
         Language = self.apps.get_model(APP_NAME, "Language")
 
-        english_test_for_adults = EnrollmentTest(
-            language=Language.objects.get(id="en"),
-        )
-
-        english_test_for_adults.save()
-        english_test_for_adults.age_ranges.set(
-            AgeRange.objects.filter(type=AgeRangeType.STUDENT, age_from__gte=18)
-        )
-
         path = (
             settings.BASE_DIR
-            / "api"
+            / APP_NAME
             / "migrations"
             / "data_migration_sources"
             / "enrollment_test_en.yaml"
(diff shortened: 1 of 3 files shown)


api/migrations/0002_data_migration.py
```diff
@@ -97,6 +99,66 @@ def _write_languages_and_levels(self):
         )
         LanguageAndLevel.objects.bulk_create(language_and_level_objects)
 
+    def _write_enrollment_tests(self):
+        """Writes `EnrollmentTest', `Question`, and `EnrollmentTestQuestionOption` to database."""
```
Roxe322, line 103: Something wrong with quotes here :)

api/migrations/0002_data_migration.py
```diff
@@ -97,6 +99,66 @@ def _write_languages_and_levels(self):
...
+                    enrollment_test=enrollment_test, text=item["text"]
+                )
+
+                # make sure exactly one option is marked as correct in each question
+                assert (
+                    len([o for o in item["options"] if o["is_correct"] is True]) == 1
+                ), f"Exactly 1 option has to be marked as correct for {item=}"
+
+                for option_data in item["options"]:
+                    EnrollmentTestQuestionOption.objects.create(
```
Roxe322, line 149: How about bulk_create?

api/migrations/0002_data_migration.py
```diff
@@ -97,6 +99,66 @@ def _write_languages_and_levels(self):
...
+            / "migrations"
+            / "data_migration_sources"
+            / "enrollment_tests.yaml"
+        )
+        with path.open(encoding="utf8") as fh:
+            data = yaml.safe_load(fh)
+
+        for block in data:
+            enrollment_test = EnrollmentTest.objects.create(
+                language=Language.objects.get(id=block["language"]),
```
Roxe322, line 125: I think we can skip object getting by setting language_id directly

api/migrations/0002_data_migration.py
lemontree210: Yes, fixed

api/migrations/0002_data_migration.py
lemontree210: Yes :)

api/migrations/0002_data_migration.py
lemontree210: Right, since id=locale for Language. Fixed

api/migrations/0002_data_migration.py
```diff
@@ -97,6 +99,66 @@ def _write_languages_and_levels(self):
...
+                    )
+                    for option_data in item["options"]
+                ] + [  # add "don't know" to each question
+                    EnrollmentTestQuestionOption(
+                        question=question,
+                        text="(I don't know 😕)",
+                        is_correct=False,
+                    )
+                ]
+                EnrollmentTestQuestionOption.objects.bulk_create(options)
```
Roxe322, line 160: I think you can move this outside the for-loop, to be called only once. Also I'm not sure about question, but maybe it's also could be put into list for future bulk_create.

Roxe322 (approved):

api/migrations/0002_data_migration.py
lemontree210: Done for options (don't think can be done for questions)

PR #73, opened 2023-05-05
add simple view (with API docs) to check existence of Telegram chat ID

### LINKED ISSUE #61: Create missing API views for bot
- check existence of `registration_telegram_bot_chat_id`

lemontree210:
closes #61


api/views/personal_info.py
```diff
@@ -19,8 +21,40 @@ def get_serializer_class(self) -> type[BaseSerializer[PersonalInfo]]:
...
+            200: OpenApiResponse(description="Chat ID is found"),
+            404: OpenApiResponse(description="Chat ID not found"),
+        },
+    )
+    @action(detail=False, methods=["get"])
+    def check_existence_of_chat_id(self, request: Request) -> Response:
+        """
+        Check if PersonalInfo with given ``registration_telegram_bot_chat_id`` exists.
+        """
+        # This method is used without a serializer because this seems to be overkill
```
Roxe322, line 47: I think this entire comment could be removed, it's too long and not needed in my opinion

api/views/personal_info.py
```diff
@@ -19,8 +21,40 @@ def get_serializer_class(self) -> type[BaseSerializer[PersonalInfo]]:
...
+        # for a simple GET request with a single parameter.  Since name and e-mail are required
+        # fields for PersonalInfo (and we only have chat ID), we cannot use the same technique
+        # as in .check_existence(). Note that .check_existence() above returns 200 OK if the
+        # personal info does NOT exist and 400 BAD REQUEST if it DOES exist.
+        # This is because POST method is used up there, so non-existent user can be "created",
+        # hence success is reported if no user was found.
+        # Here we use GET and the responses are different.
+
+        if PersonalInfo.objects.filter(
+            registration_telegram_bot_chat_id=request.GET.get("chat_id")
```
Roxe322, line 57: If I will call this method without "chat_id" param - it will return 200 if we have PersonalInfo with  registration_telegram_bot_chat_id=None.

I think we should make this param required, and then it's better to put this checking logic into serializer

lemontree210: Well actually it makes absolutely no sense to call this view without chat ID (a bot will always have some ID), but you're right in that we should make an implementation that doesn't care about who is using it.  

Maybe I don't understand serializers, but how can we use it if we're in the very beginning and haven't even tried to create a `PersonalInfo`, we have no name, no email yet (and they are required for `PersonalInfo`).

For now, I just added the logic for handling the missing `chat_id`. Maybe that will do :)

Roxe322: Serializer is not required to be ModelSerializer. It could be just description of some fields with `validate` method. And here in view you should just pass request.query_params into it, as data.

Roxe322: I mean just regular Serializer class.

https://www.django-rest-framework.org/tutorial/1-serialization/
https://www.django-rest-framework.org/api-guide/serializers/

lemontree210: Done

api/serializers/personal_info.py
```diff
@@ -21,3 +27,24 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
...
+class CheckChatIdExistenceSerializer(serializers.ModelSerializer[PersonalInfo]):
+    """A serializer used to check if chat ID from telegram registration bot exists in database.
+
+    Since it is impossible to create a `PersonalInfo` item without a name and an email,
+    this serializer is supposed to be used with GET requests.
+    """
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        # self.fields.keys() keeps mypy happy, but ruff will remove .keys() and make mypy complain
+        if not all(f in attrs for f in self.fields.fields):
```
Roxe322, line 41: Just mark needed field(s) as required and this will be checked by serializer. 
https://stackoverflow.com/a/39835680

lemontree210: Done

api/views/personal_info.py
```diff
@@ -1,26 +1,50 @@
...
         """
         serializer = self.get_serializer(data=request.data)
         serializer.is_valid(raise_exception=True)
         return Response(status.HTTP_200_OK)
+
+    @extend_schema(
+        parameters=[
+            OpenApiParameter(name="registration_telegram_bot_chat_id", type=int, required=True),
+        ],
+    )
```
Roxe322, line 39:
```suggestion
    @extend_schema(
        parameters=[CheckChatIdExistenceSerializer],
    )
```

lemontree210: Done

Roxe322 (approved):

PR #75, opened 2023-05-06
fix(`PersonalInfo`): Capitalize `first_name` and `last_name`, lowercase `email`

### LINKED ISSUE #60: Capitalize names and casefold emails in `PersonalInfo`
[Check for existence](https://github.com/samanthasgroup/django-webapps/blob/master/api/serializers/personal_info.py#L16) should be made in case insensitive mode. All strings should be `.strip`ped too.

Change the `.save()` method on `PersonalInfo` to achieve this.

lemontree210:
closes #60

### DIFF
--- a/api/models/people.py
+++ b/api/models/people.py
@@ -100,7 +100,7 @@ class PersonalInfo(GroupOrPerson):
 
         self.first_name = capitalize_each_word(self.first_name)
         self.last_name = capitalize_each_word(self.last_name)
-        self.email = self.email.casefold()
+        self.email = self.email.lower()
         super().save(
             force_insert=force_insert,
             force_update=force_update,
--- a/api/models/people.py
+++ b/api/models/people.py
@@ -1,4 +1,5 @@
 import uuid
+from collections.abc import Iterable
 from datetime import timedelta
 
 from django.db import models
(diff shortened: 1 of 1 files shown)


api/models/people.py
```diff
@@ -87,6 +88,26 @@ class Meta:
...
+        force_update: bool = False,
+        using: str | None = None,
+        update_fields: Iterable[str] | None = None,
+    ) -> None:
+        def capitalize_each_word(str_: str) -> str:
+            return " ".join(part.capitalize() for part in str_.split())
+
+        self.first_name = capitalize_each_word(self.first_name)
+        self.last_name = capitalize_each_word(self.last_name)
+        self.email = self.email.casefold()
```
Roxe322, line 103: Could you please explain, why not `.lower()`?

I've just known about such method existence, never seen it's usage before

lemontree210: A good explanation is e.g. [here](https://stackoverflow.com/a/45745761). I actually think I should use `.lower()` in this case because we are persisting, not comparing. I was thinking of comparisons (data sent by bot vs. data already in database) and used `.casefold()`, but in the context of `.save()` it's probably wrong.

lemontree210, line 103:
```suggestion
        self.email = self.email.lower()
```

Roxe322 (approved):

PR #81, opened 2023-05-13
All teachers without personal info view

Roxe322:


api/models/choices/non_teaching_help.py
lemontree210, line 1: I think you should then also include these choices into `models.NonTeachingHelp`.

api/models/constants.py
```diff
@@ -90,3 +90,16 @@ class CoordinatorGroupLimit(IntEnum):
...
+    "can_check_syllabus",
+    "can_host_mentoring_sessions",
+    "can_give_feedback",
+    "can_help_with_childrens_groups",
+    "can_provide_materials",
+    "can_invite_to_class",
+    "can_work_in_tandem",
+]
+
+TEACHER_PEER_SUPPORT_FIELD_NAME_PREFIX = "peer_support_"
```
lemontree210, line 105: It makes sense to switch these constants in code: both alphabetically and logically

api/schema.py
```diff
@@ -33,3 +38,41 @@ def _get_response_bodies(self, direction: str = "response") -> dict[str, Any]:
                 status_code="400",
             )
         return bodies
+
+
+class TeacherNonTeachingHelpProvidedFieldFix(OpenApiSerializerFieldExtension):
```
lemontree210, line 43: Maybe this will change or not be needed at all if we add the choices to the model itself?

api/serializers/non_teaching_help.py
```diff
@@ -1,9 +1,29 @@
...
+from api.models.choices.non_teaching_help import NonTeachingHelpType
 
 
 class NonTeachingHelpSerializer(serializers.ModelSerializer[NonTeachingHelp]):
     class Meta:
         model = NonTeachingHelp
         fields = "__all__"
+
+
+class NonTeachingHelpPublicSerializerField(serializers.Field):  # type: ignore
```
lemontree210, line 14: Not sure "Public" is needed here. Just listing types of help should be the same in "public" and "private" views. Either both non-teaching help and peer help fields should be called "public" or none of them IMO

lemontree210 (changes requested):
Thanks for the PR! I guess the only major question is about adding `NonTeachingHelpType` as choices to `NonTeachingHelp` model directly.

api/schema.py
Roxe322: I don't think so. Choices can help in schema when they show some validation method - to show only some values as acceptable. But here we show all this fields as nested field, so we need to fix schema maually

api/serializers/non_teaching_help.py
Roxe322: Removed Public here :)

api/models/constants.py
Roxe322: Done

api/models/choices/non_teaching_help.py
Roxe322: Yeah, forgot about it. Done

api/schema.py
lemontree210: Yes, I figured as much, just wanted to check.

api/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 4.2 on 2023-05-01 12:07
+# Generated by Django 4.1.7 on 2023-05-13 19:26
```
lemontree210, line 1: Huh? :) `poetry install`? :)
Not that it really matters, of course 😆

api/migrations/0003_fake_data.py
```diff
@@ -60,6 +60,7 @@ class RecipeStorage:
 
     def __init__(self):
         self.faker: Faker = Faker()
+        Faker.seed(0)
```
lemontree210, line 63: Does this make sure the random names are always the same?

api/models/choices/non_teaching_help.py
lemontree210, line 1: I would also suggest renaming it to `non_teaching_help_type.py` so that name of module matches name of class (like with `age_range_type.py`)

api/schema.py
```diff
@@ -33,3 +38,41 @@ def _get_response_bodies(self, direction: str = "response") -> dict[str, Any]:
...
+    """
+    This class automatically extends schema for NonTeachingHelpSerializerField,
+    showing this object properly in docs.
+    """
+
+    target_class = "api.serializers.non_teaching_help.NonTeachingHelpSerializerField"
+
+    def map_serializer_field(
+        self, auto_schema: AutoSchema, direction: Literal["request", "response"]  # noqa: ARG002
```
lemontree210, line 52: If you're really into avoiding repetitions, you could create 
```py
class Direction = Literal["request", "response"]
```
but it's totally up to you :)

api/schema.py
```diff
@@ -33,3 +38,41 @@ def _get_response_bodies(self, direction: str = "response") -> dict[str, Any]:
                 status_code="400",
             )
         return bodies
+
+
+class TeacherNonTeachingHelpProvidedFieldFix(OpenApiSerializerFieldExtension):
+    """
+    This class automatically extends schema for NonTeachingHelpSerializerField,
+    showing this object properly in docs.
```
lemontree210, line 46:
```suggestion
    Extends schema for `NonTeachingHelpSerializerField` for proper representation in docs.
```
So it fits on one line and we don't have to follow the convention with summary line and the rest on other lines :)
Same for  `TeacherPeerSupportFieldFix` class below.

lemontree210 (changes requested):
Only nitpicks, most of them you can just ignore if you wish - I guess except renaming the module for `NonTeachingHelpType` :)

api/migrations/0003_fake_data.py
Roxe322: To be honest, I'm not sure how it works :D But hope it will help to prevent problems of constraints violating.

As I understand - during every "clear" recreation it will produce same values, but of course in one "session" all values would be random.

api/migrations/0001_initial.py
Roxe322: Fixed it back to 4.2 and updated my local version :)

api/models/choices/non_teaching_help.py
Roxe322: done

api/schema.py
Roxe322: I think without `class` :) Done

api/schema.py
Roxe322: api/schema.py:46:100: E501 Line too long (101 > 99 characters)

Roxe322: I think this convention should not to be followed if it doesn't fit into one-line, but it should be a one-line by design

Roxe322: But I still reworded it here and below

api/schema.py
lemontree210: Yes, of course 🤦🏻

api/schema.py
lemontree210: > I think this convention should not to be followed if it doesn't fit into one-line, but it should be a one-line by design

OK :)

lemontree210 (approved):

PR #82, opened 2023-05-14
suppress `DatabaseError` correctly when generating fake data

### LINKED ISSUE #64: Fake data migration fails sometimes because of constraints in models
Known cases, but there may be more:

- `UNIQUE constraint failed: api_group.telegram_chat_url`
- `CHECK constraint failed: at_least_one_day_time_slot_must_be_selected`

lemontree210:
closes #64

### DIFF
--- a/api/migrations/0003_fake_data.py
+++ b/api/migrations/0003_fake_data.py
@@ -1,7 +1,7 @@
 import sys
 from datetime import timedelta, time
 
-from django.db import migrations, models, transaction, IntegrityError
+from django.db import migrations, models, transaction, DatabaseError
 from django.db.backends.sqlite3.schema import DatabaseSchemaEditor
 from django.db.migrations.state import StateApps
 from django.db.models import Q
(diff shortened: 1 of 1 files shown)


api/migrations/0003_fake_data.py
```diff
@@ -278,11 +278,16 @@ def _make_amount_of_recipe_and_skip_database_error(
...
-                    break
-                except DatabaseError:
-                    pass
+                    # context manager has to be used to avoid TransactionManagementError.
+                    # See section "Controlling transactions explicitly" in
+                    # https://docs.djangoproject.com/en/dev/topics/db/transactions/
+                    with transaction.atomic():
+                        recipe.make()
+                        break
+                except IntegrityError as e:
```
Roxe322, line 289: IntegrityError is a subclass of DatabaseError, so I think in previous version it also should work 🤔

lemontree210: OK, but the main thing is that I added `transaction.atomic()`

Roxe322: Ah, ok. But probably let's keep DatabaseError to avoid any other db exception?

lemontree210: Done

Roxe322 (approved):

PR #88, opened 2023-05-19
Return result of enrollment test (level) without writing the answers to database

### LINKED ISSUE #89: Get results of enrollment test without writing answers to database
In the current implementation, backend returns level of English as a response to a POST request creating records for `EnrollmentTestAnswer` linked to the given student. However, when the student is taken an assessment, they are still in the middle of the registration process, and their record will not be created for a while after that (they need to answer some additional questions or even take another assessment). It even can't be created at that stage because some required fields of `Student` cannot be filled yet.

That means we need a view that just gets IDs of `EnrollmentTestAnswer` and returns level, nothing more.

The existing view will still be used to create records for the answers later on (maybe without returning the level as this won't be needed anymore).

lemontree210:
closes #89

### DIFF
--- a/api/migrations/0002_data_migration.py
+++ b/api/migrations/0002_data_migration.py
@@ -9,7 +9,7 @@ from api.models.choices.non_teaching_help_type import NonTeachingHelpType
 from api.models.constants import (
     STUDENT_AGE_RANGES,
     STUDENT_AGE_RANGES_FOR_MATCHING,
-    STUDENT_AGE_RANGES_FOR_TEACHER,
+    STUDENT_AGE_RANGES_FOR_TEACHER, LanguageLevelId,
 )
 from api.models.data_populator import DataPopulator
 
@@ -89,8 +89,16 @@ class InitialDataPopulator(DataPopulator):
         )
         Language.objects.bulk_create(languages)
 
-        # no C2 at this school
-        levels = (Level(id=id_) for id_ in ("A0", "A1", "A2", "B1", "B2", "C1"))
+        levels = (
+            Level(id=id_) for id_ in (
+            LanguageLevelId.A0_BEGINNER,
+            LanguageLevelId.A1_ELEMENTARY,
+            LanguageLevelId.A2_PRE_INTERMEDIATE,
+            LanguageLevelId.B1_INTERMEDIATE,
+            LanguageLevelId.B2_UPPER_INTERMEDIATE,
+            LanguageLevelId.C1_PRE_ADVANCED
+            )
+        )
         Level.objects.bulk_create(levels)
 
         # English is taught at all levels, all other languages at levels A0 through A2
--- a/api/migrations/0002_data_migration.py
+++ b/api/migrations/0002_data_migration.py
@@ -89,16 +89,7 @@ class InitialDataPopulator(DataPopulator):
         )
         Language.objects.bulk_create(languages)
 
-        levels = (
-            Level(id=id_) for id_ in (
-            LanguageLevelId.A0_BEGINNER,
-            LanguageLevelId.A1_ELEMENTARY,
-            LanguageLevelId.A2_PRE_INTERMEDIATE,
-            LanguageLevelId.B1_INTERMEDIATE,
-            LanguageLevelId.B2_UPPER_INTERMEDIATE,
-            LanguageLevelId.C1_PRE_ADVANCED
-            )
-        )
+        levels = (Level(id=id_) for id_ in LanguageLevelId)
         Level.objects.bulk_create(levels)
 
         # English is taught at all levels, all other languages at levels A0 through A2
--- a/api/models/enrollment_tests.py
+++ b/api/models/enrollment_tests.py
@@ -92,10 +92,7 @@ class EnrollmentTestResult(models.Model):
 
     @staticmethod
     def calculate_level(answer_ids: Sequence[int] | QuerySet[Any]) -> str:
-        """Calculates language level depending on amount of correct answers.
-
-        This method can be called from outside without creating any `EnrollmentTestResult` objects.
-        """
+        """Calculates language level depending on amount of correct answers."""
         # depending on number of questions in test, the thresholds are different
         total_answers = len(answer_ids)
 
--- a/api/models/enrollment_tests.py
+++ b/api/models/enrollment_tests.py
@@ -1,3 +1,5 @@
+from collections.abc import Sequence
+
 from django.db import models
 
 from api.models.age_ranges import AgeRange
@@ -81,13 +83,20 @@ class EnrollmentTestResult(models.Model):
     @property
     def resulting_level(self) -> str:
         """Returns language level of the student based on amount of correct answers."""
+        return self.calculate_level([answer.id for answer in self.answers.all()])
 
+    @staticmethod
+    def calculate_level(answer_ids: Sequence[int]) -> str:
+        """Calculates language level depending on amount of correct answers.
+
+        This is a static method that can be called from outside without creating any records.
+        """
         # depending on number of questions in test, the thresholds are different
         thresholds_for_number_of_questions = {
             25: {6: "A1", 11: "A2", 19: "B1"},
             35: {6: "A1", 13: "A2", 20: "B1", 27: "B2", 32: "C1"},
         }
-        total_answers = self.answers.count()
+        total_answers = len(answer_ids)
 
         # All answers must be filled, so it is safe to check number of answers, not questions
         if total_answers not in thresholds_for_number_of_questions:
@@ -95,7 +104,8 @@ class EnrollmentTestResult(models.Model):
                 f"Enrollment test with {total_answers} questions is not supported"
             )
 
-        number_of_correct_answers = self.answers.filter(is_correct=True).count()
+        answers = EnrollmentTestQuestionOption.objects.filter(id__in=answer_ids)
+        number_of_correct_answers = answers.filter(is_correct=True).count()
         level = "A0"
         for threshold in thresholds_for_number_of_questions[total_answers]:
             if number_of_correct_answers >= threshold:
--- a/api/models/enrollment_tests.py
+++ b/api/models/enrollment_tests.py
@@ -78,6 +78,10 @@ class EnrollmentTestResult(models.Model):
     def __str__(self) -> str:
         return (
             f"Answers to enrollment test by {self.student} "
-            f"({self.answers.filter(is_correct=True).count()} correct "
+            f"({self.correct_answers_count} correct "
             f"out of {self.answers.count()})"
         )
+
+    @property
+    def correct_answers_count(self) -> int:
+        return self.answers.filter(is_correct=True).count()
--- a/api/models/enrollment_tests.py
+++ b/api/models/enrollment_tests.py
@@ -1,7 +1,7 @@
 from django.db import models
 
 from api.models.age_ranges import AgeRange
-from api.models.constants import DEFAULT_CHAR_FIELD_MAX_LEN, LEVEL_BY_PERCENTAGE_OF_CORRECT_ANSWERS
+from api.models.constants import DEFAULT_CHAR_FIELD_MAX_LEN
 from api.models.languages_levels import Language
 from api.models.people import Student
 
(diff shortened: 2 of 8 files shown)


api/views/enrollment_test.py
```diff
@@ -16,4 +24,20 @@
...
+    def get_level(self, request: Request) -> Response:
+        """Calculates level of language based on number of correct answers."""
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        answer_ids = cast(list[int], request.POST.getlist("answers"))
+        try:
+            return Response(EnrollmentTestResultLevelSerializer.calculate_level(answer_ids))
+        except NotImplementedError as e:
+            raise ValidationError(str(e))
```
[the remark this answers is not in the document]
lemontree210: Fixed

api/models/enrollment_tests.py
```diff
@@ -81,11 +83,34 @@ def __str__(self) -> str:
     @property
     def resulting_level(self) -> str:
         """Returns language level of the student based on amount of correct answers."""
-        correct_answers = self.answers.filter(is_correct=True).count()
-        total_answers = self.answers.count()
-        percentage = correct_answers / total_answers * 100
-        closest_percentage = min(
-            LEVEL_BY_PERCENTAGE_OF_CORRECT_ANSWERS, key=lambda x: abs(x - percentage)
-        )
+        return self.calculate_level([answer.id for answer in self.answers.all()])
```
Roxe322, line 86: self.answers.values_list('id', flat=True)

api/models/enrollment_tests.py
```diff
@@ -81,11 +83,34 @@ def __str__(self) -> str:
...
+
+    @staticmethod
+    def calculate_level(answer_ids: Sequence[int]) -> str:
+        """Calculates language level depending on amount of correct answers.
+
+        This method can be called from outside without creating any `EnrollmentTestResult` objects.
+        """
+        # depending on number of questions in test, the thresholds are different
+        thresholds_for_number_of_questions = {
```
Roxe322, line 95: Maybe let's save it as a constant outside the function?

api/views/enrollment_test.py
```diff
@@ -16,4 +24,24 @@ class EnrollmentTestViewSet(ReadOnlyModelViewSet[EnrollmentTest]):
...
+            return EnrollmentTestResultLevelSerializer
+        return EnrollmentTestResultCreateSerializer
+
+    @action(detail=False, methods=["post"])
+    def get_level(self, request: Request) -> Response:
+        """Calculates level of language based on number of correct answers."""
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        answer_ids = cast(list[int], request.POST.getlist("answers"))
```
Roxe322, line 39: Why don't you use serializer.validated_data here?

api/views/enrollment_test.py
```diff
@@ -16,4 +24,24 @@ class EnrollmentTestViewSet(ReadOnlyModelViewSet[EnrollmentTest]):
...
+        try:
+            return Response(EnrollmentTestResultLevelSerializer.calculate_level(answer_ids))
+        except NotImplementedError:
+            # Repeating the message rather than putting str(e), otherwise CodeQL complains about
```
Roxe322, line 43: I think we could remove this comment, it's pretty clear why we do any rewording here

api/models/enrollment_tests.py
```diff
@@ -81,11 +83,34 @@ def __str__(self) -> str:
...
-        closest_percentage = min(
-            LEVEL_BY_PERCENTAGE_OF_CORRECT_ANSWERS, key=lambda x: abs(x - percentage)
-        )
+        return self.calculate_level([answer.id for answer in self.answers.all()])
+
+    @staticmethod
+    def calculate_level(answer_ids: Sequence[int]) -> str:
+        """Calculates language level depending on amount of correct answers.
+
+        This method can be called from outside without creating any `EnrollmentTestResult` objects.
```
Roxe322, line 92: It's clear because it's staticmethod :)

api/models/enrollment_tests.py
```diff
@@ -81,11 +83,34 @@ def __str__(self) -> str:
...
+        """
+        # depending on number of questions in test, the thresholds are different
+        thresholds_for_number_of_questions = {
+            25: {5: "A1", 11: "A2", 19: "B1"},
```
Roxe322, line 96: Also I really think we should make an enum with levels ids

api/serializers/enrollment_test.py
```diff
@@ -50,3 +52,16 @@ class Meta:
 
     # TODO Think about some validation, e.g. answers should be unique for each question
     #  and should be in options for this question
+
+
+class EnrollmentTestResultLevelSerializer(serializers.ModelSerializer[EnrollmentTestResult]):
+    """A serializer used to get level of language based on how many answers are correct."""
+
+    @staticmethod
+    def calculate_level(answer_ids: Sequence[int]) -> dict[str, str]:
```
Roxe322, line 61: I think it should be SerializerMethodField. This field will be read-only, and `answers` should be write-only, so in view you will be able to general staff like `validate` and then `return serializer.data`

Roxe322 (changes requested):

api/models/enrollment_tests.py
lemontree210: Done. Yes, I'm not that familiar with QuerySet API yet. And damn how docstrings are sparse in Django!

api/models/enrollment_tests.py
lemontree210: Done

api/views/enrollment_test.py
lemontree210: Done

api/models/enrollment_tests.py
lemontree210: Removed

api/models/enrollment_tests.py
lemontree210: Done

api/views/enrollment_test.py
lemontree210: Because I forgot this thing existed :)
Done

api/serializers/enrollment_test.py
lemontree210: Fixed, I think. Wow that took some effort :)

api/migrations/0002_data_migration.py
```diff
@@ -89,8 +89,16 @@ def _write_languages_and_levels(self):
         )
         Language.objects.bulk_create(languages)
 
-        # no C2 at this school
-        levels = (Level(id=id_) for id_ in ("A0", "A1", "A2", "B1", "B2", "C1"))
+        levels = (
+            Level(id=id_) for id_ in (
```
Roxe322, line 93: You are able to iterate just through enum class LanguageLevelId.

`Level(id=id_) for id_ in LanguageLevelId`

api/models/enrollment_tests.py
```diff
@@ -76,16 +76,8 @@ class EnrollmentTestResult(models.Model):
...
-    def resulting_level(self) -> str:
-        """Returns language level of the student based on amount of correct answers."""
-        correct_answers = self.answers.filter(is_correct=True).count()
-        total_answers = self.answers.count()
-        percentage = correct_answers / total_answers * 100
-        closest_percentage = min(
-            LEVEL_BY_PERCENTAGE_OF_CORRECT_ANSWERS, key=lambda x: abs(x - percentage)
+        return (
+            f"Answers to enrollment test by {self.student} "
+            f"({self.answers.filter(is_correct=True).count()} correct "
```
Roxe322, line 81: In my opinion, put any logic inside f-string is bad practice.
I could agree with `self.answers.count()`, but this one looks too complex to be inside f-string

api/serializers/enrollment_test.py
```diff
@@ -45,8 +52,43 @@ class Meta:
...
+
+
+class EnrollmentTestResultLevelSerializer(serializers.ModelSerializer[EnrollmentTestResult]):
+    """A serializer used to get level of language based on how many answers are correct."""
+
+    resulting_level = serializers.SerializerMethodField()
+
+    class Meta:
+        model = EnrollmentTestResult
+        fields = ["answers", "resulting_level"]
```
Roxe322, line 68: Just a small optimisation: if something is not subject to change, it's better to use immutable data structure (tuple in this case)

api/serializers/enrollment_test.py
```diff
@@ -45,8 +52,43 @@ class Meta:
...
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        if len(attrs["answers"]) not in ENROLLMENT_TEST_LEVEL_THRESHOLDS_FOR_NUMBER_OF_QUESTIONS:
+            raise serializers.ValidationError(
+                f"Enrollment test with {len(attrs['answers'])} questions is not supported"
```
Roxe322, line 73: `len(attrs["answers"])` used twice here, so it's better to put it into variable

api/views/enrollment_test.py
```diff
@@ -16,4 +22,21 @@ class EnrollmentTestViewSet(ReadOnlyModelViewSet[EnrollmentTest]):
...
+            return EnrollmentTestResultLevelSerializer
+        return EnrollmentTestResultCreateSerializer
+
+    @action(detail=False, methods=["post"])
+    def get_level(self, request: Request) -> Response:
+        """Calculates level of language based on number of correct answers."""
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        try:
+            return Response({"resulting_level": serializer.data["resulting_level"]})
```
Roxe322, line 37: Why not just

`return Response(serializer.data)`

?

Roxe322: If you will make field `answers` write-only, it won't appear in response

api/views/enrollment_test.py
```diff
@@ -16,4 +22,21 @@ class EnrollmentTestViewSet(ReadOnlyModelViewSet[EnrollmentTest]):
...
+        except NotImplementedError:
+            raise ValidationError(
+                f"Enrollment test with {len(serializer.validated_data['answers'])} "
```
Roxe322, line 40: I think here you could use serializer.data to make it shorter

Roxe322: Also, this excepting here doesn't look good, I think it should be excepted in serializer's validate method

api/migrations/0002_data_migration.py
lemontree210: Done

api/models/enrollment_tests.py
lemontree210: Done

api/serializers/enrollment_test.py
lemontree210: Done (with mine and yours)

api/serializers/enrollment_test.py
lemontree210: Fixed

api/views/enrollment_test.py
lemontree210: Cool! Fixed

api/views/enrollment_test.py
lemontree210: Oh, I actually should have deleted the whole `try...except` bit and just leave `return` statement. The block was there before I implemented `.validate` in the serializer. Removed.

api/serializers/enrollment_test.py
```diff
@@ -45,8 +52,47 @@ class Meta:
...
+class EnrollmentTestResultLevelSerializer(serializers.ModelSerializer[EnrollmentTestResult]):
+    """A serializer used to get level of language based on how many answers are correct."""
+
+    resulting_level = serializers.SerializerMethodField()
+
+    class Meta:
+        model = EnrollmentTestResult
+        fields = ("answers", "resulting_level")
+        extra_kwargs = {
+            "answers": {"write_only": True}  # to exclude from serializer.data in ViewSet
```
Roxe322, line 70: Let's remove this comment :)

lemontree210: Done

Roxe322 (approved):

PR #93, opened 2023-05-21
Filter groups by student and coordinator

Roxe322:


api/filters/group.py
```diff
@@ -0,0 +1,16 @@
...
+from api.models import Group
+
+
+class GroupFilter(filters.FilterSet):
+    for_coordinator_email = filters.CharFilter(
+        field_name="coordinators__personal_info__email", lookup_expr="iexact"
+    )
+    for_student_email = filters.CharFilter(
+        field_name="students__personal_info__email", lookup_expr="iexact"
+    )
```
lemontree210, line 12: hm... why do we need this? No student will have any access to our dashboards, current or planned

PR #94, opened 2023-05-22
add internal `Student.id` to response to a POST request creating `Student`

lemontree210:
This is needed for us to be able to pass answers to enrollment test back to bot.

### DIFF
--- a/api/serializers/student.py
+++ b/api/serializers/student.py
@@ -13,6 +13,10 @@ from api.serializers.utc_timedelta import UTCTimedeltaField
 
 
 class StudentWriteSerializer(serializers.ModelSerializer[Student]):
+    # For Student, we need Student.id (and not just personal_info.id) so we can pass it
+    # from the bot to the backend when we send the enrollment test.
+    internal_id = serializers.IntegerField(source="pk", read_only=True)
+
     class Meta:
         model = Student
         exclude = ("children",)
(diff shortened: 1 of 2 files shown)


api/serializers/student.py
```diff
@@ -13,6 +13,10 @@
 
 
 class StudentWriteSerializer(serializers.ModelSerializer[Student]):
+    # For Student, we need Student.id (and not just personal_info.id) so we can pass it
+    # from the bot to the backend when we send the enrollment test.
+    internal_id = serializers.IntegerField(source="pk", read_only=True)
```
Roxe322, line 18: Why not just "id"? Or even "personal_info_id" since we use it as an id. Just to be consistent and don't rename anything.

api/views/student.py
```diff
@@ -6,7 +6,11 @@
 
 
 class StudentViewSet(ReadWriteSerializersMixin, viewsets.ModelViewSet[Student]):  # type: ignore
-    """Student viewset."""
+    """ViewSet for reading and writing data from/to students' table.
+
+    After creating a student with POST, returns student's **internal ID** along with other info.
```
Roxe322, line 11: I think this comment is not needed since it is visible in the serializer

api/serializers/student.py
```diff
@@ -13,6 +13,10 @@
 
 
 class StudentWriteSerializer(serializers.ModelSerializer[Student]):
+    # For Student, we need Student.id (and not just personal_info.id) so we can pass it
```
Roxe322, line 16: As I understand Student.id is same as personal_info.id. Student doesn't have his own id.

Roxe322 (changes requested):

lemontree210:
Closing because `Student` does not have its own ID. I programmed it this way and then forgot about it 🤦🏻 . Thank you :)

PR #108, opened 2023-06-03
fix: prevent error screen for `PersonalInfo`, implement case-insensitive existence check in serializer

### LINKED ISSUE #74: Admin site should give friendly message when UniqueConstraint fails
The error emerges because we implement a custom `.save()` method for `PersonalInfo`. The case-insensitive uniqueness of `PersonalInfo` is checked there (#99). In case of a validation error, an exception is thrown. Add a `...Form` and an `...Admin` class to `admin.py` to fix this.

### LINKED ISSUE #99: Check for existence of user should be case-insensitive (for name and email)
this issue is related to #60. Despite it being fixed for the *model*, it is still possible for a *client* to send e.g. an all-lowercase name and get a false negative result, because right now names are capitalized at creation, but not during this check.

lemontree210:
fixes #74 
fixes #99

### DIFF
--- a/api/models/people.py
+++ b/api/models/people.py
@@ -22,6 +22,7 @@ from api.models.constants import (
 from api.models.days_time_slots import DayAndTimeSlot
 from api.models.languages_levels import LanguageAndLevel
 from api.models.non_teaching_help import NonTeachingHelp
+from api.utils import capitalize_each_word
 
 
 # One person can perform several roles.  Therefore, the logic proposed is as follows: first,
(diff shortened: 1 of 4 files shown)


api/models/people.py
```diff
@@ -108,6 +105,10 @@ def capitalize_each_word(str_: str) -> str:
             update_fields=update_fields,
         )
 
+    @staticmethod
+    def capitalize_each_word(str_: str) -> str:
```
Roxe322, line 109: Are you sure it should be part of PersonalInfo class? I advise to do some utils/helpers module and put it there, since it could be used somewhere else and it hasn't direct relation to PersonalInfo class

lemontree210: Done

Roxe322 (approved):

PR #109, opened 2023-06-05
implement rules for log events in statuses (1st iteration)

lemontree210:


api/models/log_events.py
```diff
@@ -64,6 +69,25 @@ def __str__(self) -> str:
...
 
+    def save(
+        self,
+        force_insert: bool = False,
+        force_update: bool = False,
+        using: str | None = None,
+        update_fields: Iterable[str] | None = None,
+    ) -> None:
+        # TODO rules needed here. This is just a quick proof of concept
+        self.coordinator.status = CoordinatorStatus.WORKING_LIMIT_REACHED
```
Roxe322, line 80: I don't think it should be put into save method. You will have a lot of rules/statuses, etc. 
So imho it should be separate class kind of `StudentEventProcessor` with methods like "move_to_group", which will encapsulate logic of such operations for students. It could be inherited from some base class (e.g. `BaseEventProcessor`) with basic operations such as `add_status(object: Model, status: BaseStatus)`

api/models/rules/rules.py
```diff
@@ -0,0 +1,38 @@
...
+)
+from api.models.choices.statuses import (
+    CoordinatorStatus,
+    GroupStatus,
+    StudentStatus,
+    TeacherStatus,
+    TeacherUnder18Status,
+)
+
+LogEventType = (
```
Roxe322, line 18: Maybe do some base class for all event types, like Protocol (https://peps.python.org/pep-0544/) ?

api/models/rules/rules.py
```diff
@@ -0,0 +1,38 @@
...
+    CoordinatorLogEventType
+    | GroupLogEventType
+    | StudentLogEventType
+    | TeacherLogEventType
+    | TeacherUnder18LogEventType
+)
+
+
+@dataclass(frozen=True)
+class Result:
```
Roxe322, line 28: I think it should be different classes for different operation results, maybe with same interface/protocol. Now it seems too complex, with some unnecessary fields in some cases.

api/serializers/log_event.py
```diff
@@ -0,0 +1,125 @@
...
+
+    class Meta:
+        model = CoordinatorLogEvent
+        fields = ("coordinator", "date_time", "type", "group", "comment")
+
+
+class CoordinatorLogEventWriteSerializer(serializers.ModelSerializer[CoordinatorLogEvent]):
+    """A serializer for writing coordinator's log events."""
+
+    coordinator_id = serializers.IntegerField(required=True)
```
Roxe322, line 30: Maybe PrimaryKeyRelatedField could be used here?

Roxe322: Same below for all `XXX_id` fields

api/models/rules/rules.py
Roxe322: But maybe I didn't get this design idea at all :D

lemontree210:
dropping in favor of solving separate issues tracked by #116

PR #119, opened 2023-06-11
create API view to start a group

### LINKED ISSUE #117: start

### LINKED ISSUE #121: BUG: `test_public_group_list()` and `..._retrieve()` don't test students, teachers and coordinators in `Group`
The group created in the test contains no coordinators, teachers or students because `m2m` is not set to `True`. Once it is set to `True`, one gets the following traceback:
```
FAILED tests/tests_api/test_group.py::test_public_group_list - AttributeError: 'Coordinator' object has no attribute 'id'
FAILED tests/tests_api/test_group.py::test_public_group_retrieve - AttributeError: 'Coordinator' object has no attribute 'id'
```

lemontree210:
closes #117 
also fixes #121


api/views/group.py
```diff
@@ -24,3 +29,19 @@ def get_serializer_class(self) -> type[BaseSerializer[Group]]:
...
+    @action(detail=True, methods=["post"])
+    def start(self, request: Request) -> Response | HttpResponse:
+        data = request.data
+        try:
+            group_id = data["id"]
+        except KeyError:
+            return HttpResponseBadRequest("Pass 'id' of the group in 'id' field.")
+
+        try:
+            group = Group.objects.get(pk=group_id)
```
Roxe322, line 42: get_object_or_404

api/views/group.py
```diff
@@ -24,3 +29,19 @@ def get_serializer_class(self) -> type[BaseSerializer[Group]]:
             return PublicGroupWithStudentsSerializer
 
         raise NotImplementedError(f"Unknown action: {self.action}")
+
+    @action(detail=True, methods=["post"])
+    def start(self, request: Request) -> Response | HttpResponse:
+        data = request.data
```
Roxe322, line 35: you can just call self.get_object() here, since it's detail=True action

api/processors/group.py
```diff
@@ -0,0 +1,25 @@
...
+
+class GroupProcessor(Processor):
+    @classmethod
+    def start(cls, group: Group) -> None:
+        cls._set_status(obj=group, status=GroupStatus.WORKING)
+
+        # FIXME things below depend on teacher's/coordinator's params, just showing general logic
+
+        for coordinator in group.coordinators.iterator():
+            cls._set_status(obj=coordinator, status=CoordinatorStatus.WORKING_OK)
```
Roxe322, line 19: Here and below: Now you do N hits to db to update N instances. This could be done by one `update` query for each related model

api/views/group.py
lemontree210: Done

api/views/group.py
lemontree210: Done

api/processors/group.py
lemontree210: Fixed

api/processors/group.py
```diff
@@ -0,0 +1,86 @@
+from datetime import datetime
```
Roxe322, line 1: When you import datetime this way, you are shadowing base datetime module, so I advise never do such import and instead of it write just `import datetime`. 

Very bad design solution in python standard library 🤦

api/processors/group.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def _set_statuses_start(cls, group: Group) -> None:
+        timestamp = datetime.now(tz=pytz.UTC)
+
+        cls._set_status(obj=group, status=GroupStatus.WORKING, status_since=timestamp)
+
+        # Resulting status of coordinator depends on how many groups they now have.
+        # We can't use "has_enough_groups" or "has_reached_group_limit" because they are properties
+        group.coordinators.filter(groups__count_lt=CoordinatorGroupLimit.MIN).update(
```
Roxe322, line 60: Will this filters used anywhere else? If so, I advise to put them to model Manager / custom queryset class, to have something like

`group.coordinators.filter_have_enough_groups().update(...)`

https://docs.djangoproject.com/en/4.2/topics/db/managers/

api/processors/group.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+
+        TeacherLogEvent.objects.bulk_create(
+            TeacherLogEvent(teacher=teacher, to_group=group, type=TeacherLogEventType.STUDY_START)
+            for teacher in group.teachers.iterator()
+        )
+
+    @classmethod
+    def _set_statuses_start(cls, group: Group) -> None:
+        timestamp = datetime.now(tz=pytz.UTC)
```
Roxe322, line 54: Maybe we can use timezone.now() since it's a little bit shorty 

https://docs.djangoproject.com/en/4.2/topics/i18n/timezones/#naive-and-aware-datetime-objects

api/processors/group.py
```diff
@@ -0,0 +1,86 @@
...
+    TeacherStatus,
+)
+from api.models.constants import CoordinatorGroupLimit
+from api.processors.base import Processor
+
+
+class GroupProcessor(Processor):
+    @classmethod
+    def start(cls, group: Group) -> None:
+        cls._create_log_events_start(group)
```
Roxe322, line 26: Let's do with transaction.atomic() here:
https://docs.djangoproject.com/en/4.2/topics/db/transactions/#controlling-transactions-explicitly

api/processors/group.py
```diff
@@ -0,0 +1,86 @@
...
+            status=StudentStatus.STUDYING,
+            status_since=timestamp,
+        )
+
+        group.teachers.filter(groups__count__lt=F("simultaneous_groups")).update(
+            status=TeacherStatus.TEACHING_ACCEPTING_MORE,
+            status_since=timestamp,
+        )
+
+        group.teachers.filter(groups__count__gte=F("simultaneous_groups")).update(
```
Roxe322, line 83: I'm also thinking about splitting this to three methods, kind of `_update_students_group_start`, `_update_teachers_group_start`, etc. 

Not sure really, but I think it will be a little bit clearer.

api/processors/group.py
Roxe322: Also, maybe you can do a method in custom queryset kind of `set_status(status: Status)`, which will do
`self.update(status=status, status_since=timezone.now())`
to avoid duplication of same params in these updates

api/processors/group.py
lemontree210: Fixed

api/processors/group.py
lemontree210: Done

api/processors/group.py
lemontree210: Done

api/models/people.py
```diff
@@ -135,6 +138,43 @@ def __str__(self) -> str:
         return f"{self.personal_info.full_name}. Status: {getattr(self, 'status')}"
 
 
+# TODO the module is getting big. We can move Coordinator... etc into separate modules,
+#  which will also allow to rename all model modules into singular (because "people.py"
+#  will only contain `Person`).
+class CoordinatorQuerySet(models.QuerySet["Coordinator"]):
+    def annotate_with_count(self) -> "CoordinatorQuerySet":
```
Roxe322, line 145: Let's rename it to `annotate_with_group_count`

tests/tests_api/test_group.py
```diff
@@ -77,24 +87,75 @@ def test_public_group_retrieve(api_client):
...
         ],
         "is_for_staff_only": group.is_for_staff_only,
     }
+
+
+class TestPublicGroupStart:
+    def test_public_group_start_general_check(self, api_client):
+        group = baker.make(Group, _fill_optional=True, make_m2m=True)
+
+        with transaction.atomic():
```
Roxe322, line 117: You don't need atomic here, because .save() is called once and it will be already in one transaction

tests/tests_api/test_group.py
```diff
@@ -77,24 +87,75 @@ def test_public_group_retrieve(api_client):
...
+            group.save()
+            group.coordinators.update(status=CoordinatorStatus.WORKING_BELOW_THRESHOLD)
+            group.students.update(status=StudentStatus.AWAITING_START)
+            group.teachers.update(status=TeacherStatus.AWAITING_START)
+
+        response = api_client.post(self._make_url(group))
+
+        assert response.status_code == status.HTTP_201_CREATED
+
+        group = Group.objects.get(pk=group.id)  # reload, or else status will be the old one
```
Roxe322, line 130: group.refresh_from_db()

tests/tests_api/test_group.py
```diff
@@ -77,24 +87,75 @@ def test_public_group_retrieve(api_client):
...
+        assert group.status == GroupStatus.WORKING
+
+        common_status_since = group.status_since
+        assert common_status_since.day == timezone.now().day
+        assert common_status_since.hour == timezone.now().hour
+        assert common_status_since.minute == timezone.now().minute
```
Roxe322, line 136: timezone.now() called three times, in theory it could lead to some problems. Also for such cases it's better to use freezegun (https://github.com/spulec/freezegun)

Roxe322: Not sure we need it right now, but it's definitely better to call timezone.now() only once

tests/tests_api/test_group.py
```diff
@@ -77,24 +87,75 @@ def test_public_group_retrieve(api_client):
...
+            assert teacher.status in (
+                TeacherStatus.TEACHING_ACCEPTING_MORE,
+                TeacherStatus.TEACHING_NOT_ACCEPTING_MORE,
+            )
+            assert teacher.status_since == common_status_since
+
+        # TODO test creation of log events
+
+    @staticmethod
+    def _make_url(group: Group) -> str:
```
Roxe322, line 160: There is `reverse()` in django for it :) 
https://docs.djangoproject.com/en/4.2/ref/urlresolvers/#reverse

api/models/people.py
```diff
@@ -135,6 +138,43 @@ def __str__(self) -> str:
...
+        return self.annotate(group_count=Count("groups"))
+
+    def below_threshold(self) -> "CoordinatorQuerySet":
+        """QuerySet with coordinators with not enough groups."""
+        return self.annotate_with_count().filter(group_count__lt=CoordinatorGroupLimit.MIN)
+
+    def ok(self) -> "CoordinatorQuerySet":
```
Roxe322, line 152: Are you sure about naming here? This `ok` isn't very informative imho

api/models/people.py
```diff
@@ -135,6 +138,43 @@ def __str__(self) -> str:
...
+    def limit_reached(self) -> "CoordinatorQuerySet":
+        """QuerySet with coordinators that have exceeded the limit of groups."""
+        return self.annotate_with_count().filter(group_count__gte=CoordinatorGroupLimit.MAX)
+
+
+class CoordinatorManager(models.Manager["Coordinator"]):
+    def get_queryset(self) -> CoordinatorQuerySet:
+        return CoordinatorQuerySet(self.model, using=self._db)
+
+    def below_threshold(self) -> CoordinatorQuerySet:
```
Roxe322, line 168: Afaik, you are able to do smth like


```
class SomeModel:
    objects = SomeModelQuerySet.as_manager()
```

And then you don't need to duplicate these methods

Roxe322: Same for all models

api/processors/group.py
```diff
@@ -0,0 +1,93 @@
...
+            for teacher in group.teachers.iterator()
+        )
+
+    @classmethod
+    def _set_statuses_start(cls, group: Group) -> None:
+        timestamp = timezone.now()
+
+        cls._set_status(obj=group, status=GroupStatus.WORKING, status_since=timestamp)
+
+        for category in ("coordinators", "students", "teachers"):
```
Roxe322, line 59: Are you sure it's more readable than

```
set_coordinators_status_start(group=group, timestamp=timestamp)
set_teachers_status_start(group=group, timestamp=timestamp)
set_students_status_start(group=group, timestamp=timestamp)
```

api/models/people.py
```diff
@@ -135,6 +138,43 @@ def __str__(self) -> str:
...
+        return self.get_queryset().below_threshold()
+
+    def ok(self) -> CoordinatorQuerySet:
+        return self.get_queryset().ok()
+
+    def limit_reached(self) -> CoordinatorQuerySet:
```
Roxe322, line 174: also, I think that all methods of Queryset which are filtering something should start with `filter` like
`filter_limit_reached`

Roxe322: Because QS methods can do updates or anything else, not just filtering

api/processors/group.py
Roxe322: Also, we still have these timestamps passed to every call. Do we really want all of them to be same?

api/models/people.py
lemontree210: Done

tests/tests_api/test_group.py
lemontree210: Done

tests/tests_api/test_group.py
lemontree210: Done

tests/tests_api/test_group.py
lemontree210: Done

tests/tests_api/test_group.py
lemontree210: Fixed (I still decided to leave the `_make_url()` method, though

api/models/people.py
lemontree210: Done

tests/tests_api/test_group.py
lemontree210: I'll keep `freezegun` in mind, for now I think it's OK if I just test the minutes and the hours (especially because I explicitly set `status_since` to yesterday before running `start`)

api/processors/group.py
lemontree210: > set_coordinators_status_start(group=group, timestamp=timestamp)
> set_teachers_status_start(group=group, timestamp=timestamp)
> set_students_status_start(group=group, timestamp=timestamp)

Fixed

tests/tests_api/test_group.py
```diff
@@ -77,24 +101,180 @@ def test_public_group_retrieve(api_client):
...
+        ],
+    )
+    def test_public_group_start_coordinator_status(
+        self, api_client, timestamp, number_of_groups_to_start, expected_status
+    ):
+        coordinator = baker.make(Coordinator, _fill_optional=True)
+        coordinator.status = CoordinatorStatus.WORKING_BELOW_THRESHOLD
+        coordinator.save()
+
+        for i in range(number_of_groups_to_start):
```
Roxe322, line 204: It's better to use `_` for unused variables

tests/tests_api/test_group.py
```diff
@@ -77,24 +101,180 @@ def test_public_group_retrieve(api_client):
...
+        student.refresh_from_db()
+        assert student.groups.count() == 1
+        assert student.status == StudentStatus.STUDYING
+        self._compare_date_time_with_timestamp(student.status_since, timestamp)
+
+    @staticmethod
+    def _compare_date_time_with_timestamp(
+        date_time: datetime.datetime, timestamp: datetime.datetime
+    ):
+        assert date_time.year == timestamp.year
```
Roxe322, line 242: Why don't you compare objects directly?

tests/tests_api/test_group.py
lemontree210: Done

tests/tests_api/test_group.py
lemontree210: Because test `timestamp` that is generated in `conftest.py` will never be exactly the same as the timestamp generated in the group processor. That's why I don't even check the seconds (just in case). Furthermore, when log events are generated, current time is added to them automatically, as defined in `LogEvent`:
```py
date_time = models.DateTimeField(auto_now_add=True)
```
So direct comparison will always fail.

Roxe322 (approved):

PR #136, opened 2023-06-21
Use 409 instead of 400 in personal_info/check_existence/

### LINKED ISSUE #134: Use status code 409 instead of 400 where appropriate
Status code [409 (Conflict)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409).

If we send request to check the existence of user with given name and email, and the email is deemed invalid by Django validator, we get request code `400`. For example, email address `d@f.a` passes the regex in the bot but is rejected by Django. We get `400`:
```
response.status_code=400, response.json()={'email': ['Enter a valid email address.']}
```
We should discern this situation from the one where the request itself is correct, but the user with the data really exists. I suggest using `409` for this.

Roxe322:
Closes #134

### DIFF
--- a/api/serializers/errors.py
+++ b/api/serializers/errors.py
@@ -11,3 +11,11 @@ class ValidationErrorSerializer(serializers.Serializer[Model]):
     field1 = serializers.ListField(child=serializers.CharField())
     field2 = serializers.ListField(child=serializers.CharField())
     non_field_errors = serializers.ListField(child=serializers.CharField())
+
+
+class APIExceptionSerializer(serializers.Serializer[Model]):
+    """
+    Used only in OpenAPI schema, represents the response body for default Django API exception.
+    """
+
+    detail = serializers.CharField()
--- a/api/serializers/errors.py
+++ b/api/serializers/errors.py
@@ -13,7 +13,7 @@ class ValidationErrorSerializer(serializers.Serializer[Model]):
     non_field_errors = serializers.ListField(child=serializers.CharField())
 
 
-class APIExceptionSerializer(serializers.Serializer[Model]):
+class BaseAPIExceptionSerializer(serializers.Serializer[Model]):
     """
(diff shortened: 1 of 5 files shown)


tests/tests_api/test_personal_info.py
```diff
@@ -57,8 +57,8 @@ def test_personal_info_check_existence_returns_400_with_existing_info(api_client
             "last_name": existing_personal_info.last_name,
         },
     )
-    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
lemontree210, line 60: I think it makes sense to also add a test with a Cyrillic email, making sure that `status.HTTP_400_BAD_REQUEST` is returned.

BTW, I noticed that my RFC-something-compliant regex lets the user enter an email address with one character for domain zone (e.g. "mail@mail.**c**"), but Django rejects it as an invalid email. Just a side note.

lemontree210 (changes requested):
Thanks for the changes, it would have taken me days to figure out how to do this properly :)

api/serializers/errors.py
```diff
@@ -11,3 +11,11 @@ class ValidationErrorSerializer(serializers.Serializer[Model]):
     field1 = serializers.ListField(child=serializers.CharField())
     field2 = serializers.ListField(child=serializers.CharField())
     non_field_errors = serializers.ListField(child=serializers.CharField())
+
+
+class APIExceptionSerializer(serializers.Serializer[Model]):
```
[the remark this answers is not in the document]
Roxe322: To be fair - you are right, ValidationError is a subclass of APIException. But they have different response formats. I tried to be consistent with django naming.
Renamed it to `BaseAPIExceptionSerializer`

tests/tests_api/test_personal_info.py
Roxe322: Added test

lemontree210 (approved):

PR #156, opened 2023-07-01
Add `abort group` view

ivanBarinsky:
closes #125

PR #157, opened 2023-07-03
GroupProcessor split

### LINKED ISSUE #125: API view: abort group (finish the studies prematurely)
When a coordinator (from dashboard) sends a request stating that the group needs to be aborted (the classes end prematurely):

## Create log events

- coordinator(s): `GROUP_ABORTED` (**TODO** create)
- group: `ABORTED`
- students: `GROUP_ABORTED` (**TODO** create; also, add `GROUP_FINISHED` and remove `REGISTERED`?)
- teacher(s): `GROUP_ABORTED` (**TODO** create; also, add `GROUP_FINISHED` and remove `REGISTERED`?)

## Create/change other objects
- `Group`: move every coordinator from `.coordinators` to `.coordinators_former`, same for students and teachers (this relationship becomes available after merging #130 that closes #126). The same thing can be done the other way round (moving this group from `.groups` to `.groups_former` for each coordinator, student and teacher).

## Set statuses
- group: `ABORTED`
- Its coordinators' statuses change depending on the number of groups each of them has left:
  - if it is below threshold: `WORKING_BELOW_THRESHOLD`
  - if it is above threshold and below the limit: `WORKING_OK`
  - if it has reached the limit: `WORKING_LIMIT_REACHED`
- Its students' status changes to `AWAITING_OFFER`
- Its teachers' statuses change depending on the number of the groups they are still working with:
  - if the teacher teaches other group(s): `WORKING_ACCEPTING_MORE`
  - in an unlikely case that the teacher was teaching way above their limit of simultaneous groups and is still above this limit despite this current group having been aborted:  `WORKING_NOT_ACCEPTING_MORE`
  - if there are no groups left: `AWAITING_OFFER`

Roxe322:
a suggestion to amend #156 that is supposed to close #125


api/models/auxil/status_setter.py
```diff
@@ -0,0 +1,36 @@
...
+
+        Optionally, pass a datetime object as `status_since` to have identical timestamps
+        for multiple log events.
+        """
+        obj.status = status
+        obj.status_since = status_since or timezone.now()
+        obj.save()
+
+    @staticmethod
+    def set_coordinator_statuses(timestamp: datetime.datetime) -> None:
```
ivanBarinsky, line 25: IMO its a bit hard to figure out the purpose of this function based on its name and location in the files. Is this supposed to "refresh" coordinator statuses?

api/processors/group.py
```diff
@@ -1,184 +1,14 @@
...
-            status=TeacherStatus.TEACHING_ACCEPTING_MORE,
-            status_since=timestamp,
-        )
-
-        teachers.filter_cannot_take_more_groups().update(
-            status=TeacherStatus.TEACHING_NOT_ACCEPTING_MORE,
-            status_since=timestamp,
-        )
+    def abort(group: Group) -> None:
+        GroupAbortProcessor(group=group).process()
```
ivanBarinsky, line 14: i think its a nice approach, nice and concise and tells you where to look for implementation details 👍

api/models/auxil/status_setter.py
```diff
@@ -0,0 +1,36 @@
+import datetime
+
+from django.utils import timezone
+
+from api.models import Coordinator, Group
+from api.models.choices.status import CoordinatorStatus, Status
+from api.models.shared_abstract.person import Person
+
+
+class StatusSetter:
```
ivanBarinsky, line 10: hmm, this looks like it might suffer from the same issues as `GroupProcessor` originally did... i think an approach like that could work well with creating log events since there isnt much logic there, but with setting statuses, there are different things to take into account for each action AND for each entity type

ivanBarinsky (commented):
nice improvement in readability overall, some comments on `status_setter` 👍

api/models/auxil/status_setter.py
ivanBarinsky: on the other hand, it would be nice to have a separate `StatusSetter` for each entity type to abstract away that logic and not bulk up the processor classes

api/models/auxil/status_setter.py
lemontree210: Maybe I should call it `update_coordinator_statuses` and the class `StatusUpdater` because it basically checks the rules and applies new statuses if the criteria for a different status have been met.

api/models/auxil/status_setter.py
lemontree210: For now, we are pretty sure the coordinators' statuses will follow the same rules no matter what event is triggered. In the future, though, we will have to refine the filters so we don't update statuses of coordinators that no longer are with the project. Maybe I should create a TODO.

ivanBarinsky: ok, i think i see the thinking here 👍 my only concern is that when we get to things like putting coordinators on leave, banning them, etc, presumably the logic for that will be in a different place. i might be getting too far ahead though, it makes sense with the current state of the code and we can always refactor later

api/models/auxil/status_setter.py
ivanBarinsky: yeah that sounds good, maybe even mention coordinators being `working` in the function name to make it clear what kind of update it is

PR #166, opened 2023-07-06
fix tests generated values

Auch-Auch:
fix all bake.make with models that contain unique constraints

### DIFF
--- a/tests/tests_api/test_personal_info.py
+++ b/tests/tests_api/test_personal_info.py
@@ -8,7 +8,7 @@ TODO:
 """
 import pytest
 from django.utils.dateparse import parse_duration
-from model_bakery import baker
+from model_bakery import baker, seq
 from rest_framework import status
 
 from api.models import PersonalInfo
@@ -29,7 +29,7 @@ def test_personal_info_create(api_client, fake_personal_info_data):
 
 
 def test_personal_info_update(api_client, fake_personal_info_data):
-    personal_info = baker.make(PersonalInfo)
+    personal_info = baker.make(PersonalInfo, first_name=seq("Ivan"))
     initial_count = PersonalInfo.objects.count()
 
     response = api_client.put(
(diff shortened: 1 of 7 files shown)


tests/tests_api/test_personal_info.py
```diff
@@ -29,7 +29,7 @@ def test_personal_info_create(api_client, fake_personal_info_data):
 
 
 def test_personal_info_update(api_client, fake_personal_info_data):
-    personal_info = baker.make(PersonalInfo)
+    personal_info = baker.make(PersonalInfo, first_name=seq("Ivan"))
```
lemontree210, line 32: I would add a CSI-style comment saying that we're doing the sequence for `first_name` only because that will be enough for a `UniqueConstraint` to never be violated even if `last_name` and `email` are the same.

Also, this "Ivan" is repeated many times and is maybe worth putting into a constant (although I know it's not that common in tests)

lemontree210 (commented):
Thank you for the PR! Two general remarks:
- I guess this is still work in progress or do you want to make a separate PR for each model? If not, maybe it's best to convert this PR to draft for now
- Right now groups, students and teachers get assigned *all* availability slots. This is actually fine for the operations currently being tested, it's not relevant. But I would still say that we either have to implement some functionality to randomly select several availability slots or just add a docstring to the fixture stating that it can be used for tests where it is doesn't matter how many availability slots get assigned to an entity, as long as the `UniqueConstraint` is not violated. I think that the latter solution will be OK for now.
And then later we could add another fixture that takes all availability slots and selects several of them, either randomly or not randomly (e.g. when we test the matching algorithm).

Auch-Auch:
> Thank you for the PR! Two general remarks:
> 
> * I guess this is still work in progress or do you want to make a separate PR for each model? If not, maybe it's best to convert this PR to draft for now
> * Right now groups, students and teachers get assigned _all_ availability slots. This is actually fine for the operations currently being tested, it's not relevant. But I would still say that we either have to implement some functionality to randomly select several availability slots or just add a docstring to the fixture stating that it can be used for tests where it is doesn't matter how many availability slots get assigned to an entity, as long as the `UniqueConstraint` is not violated. I think that the latter solution will be OK for now.
>   And then later we could add another fixture that takes all availability slots and selects several of them, either randomly or not randomly (e.g. when we test the matching algorithm).

For now those are all bakery models that being used, so for example AgeRange is not used anywhere in the code. So we just should be aware of `seq` for the unique constraints in the future.

lemontree210 (approved):
Thanks for the fix!

lemontree210:
> For now we can leave it as is, it still should be valid and create a separate issue for tests refactoring.

Created #167

Issue #123, opened 2023-06-13
API view: create a group and add students and teachers to it


lemontree210:
When a coordinator creates a group in ToolJet:

## Create/change objects
- create new `Group` with following attributes passed from the dashboard: 
  - language and level
  - availability slots (no exact days and times yet)
  - lesson duration (because a teacher can determine this, but the coordinator should select some initial value based on their knowledge of the business processes: e.g., advanced groups tend to have longer classes)
  - (note that `telegram_chat_url` is **not** assigned at this point because this is just a tentative group)
- assign this `Coordinator` (identified by their email) to this `Group`
- add selected `Student`s and `Teacher`s to this `Group`

Note that the backend must know exactly who is added to the group as a student and who is added as a teacher, because in rare cases people can combine these roles (in the whole project, not in one group, of course).

> @ivanBarinsky Depending on what is doable in ToolJet, you can either create 10 fields with drop-down lists for students and 2 for teachers - or just two fields if "add more" button can be created (that adds one more field).
> @Roxe322 I don't think we will implement some limitations on number of students depending on their age, it's too complicated, and the logic is not rigid enough. Let humans and bot control this.

## Create log events

Types of `LogEvent`s to be created:

- coordinator(s): none (group is still pending, so it isn't confirmed or working, hence nothing significant has happened)
- group: `FORMED`
- students: `GROUP_OFFERED`
- teacher(s): `GROUP_OFFERED`

## Set statuses
- group: `PENDING`
- coordinator: no status change
- students: `GROUP_OFFERED`
- teachers: `GROUP_OFFERED`

Remember to update `status_since` to current time in UTC for all relevant entities.

Auch-Auch:
Noticed some schema issues:

1. `telegram_chat_url` has a unique constraint but at the same time it can not be null. So it is impossible to create Group without specifying `telegram_chat_url`.
2. I think everything with `*_former` should have a default value because during first creation group has no any `*_former` items.

lemontree210:
Thanks for the valuable insight!
>     1. `telegram_chat_url` has a unique constraint but at the same time it can not be null. So it is impossible to create Group without specifying `telegram_chat_url`.

[...]

Issue #124, opened 2023-06-15
API view: confirm the group is ready to start studies


lemontree210:
When a coordinator (from dashboard) or bot send a request stating that the group has been **confirmed by teacher(s) and students and is ready to start**:

## Create log events

Types of `LogEvent`s to be created:

- coordinator(s): `TOOK_NEW_GROUP` (`group=<this group>`)
- group: `CONFIRMED`
- students: `GROUP_CONFIRMED`  (`to_group=<this group>`)
- teacher(s): `GROUP_CONFIRMED`  (`to_group=<this group>`)

## Set statuses
- group: `AWAITING_START`
- coordinator: no change (status will change when the group starts studies, see #119)
- students: `AWAITING_START`
- teacher(s): `AWAITING_START`

Remember to update `status_since` to current time in UTC in objects (where applicable).

PR #174, opened 2023-07-08
Group creation

Auch-Auch:
PR draft for #123.

Added views for creating Groups according to project API conventions (means that Groups are not public).

For creation use POST or PUT, example of body with minimum set of fields required to create a group:

```
{
  "lesson_duration_in_minutes": 20,
  "communication_language_mode": "ru",
  "availability_slots_for_auto_matching": [2],
  "language_and_level": 1,
  "monday": "10:00:00",
  "coordinators": [
    36
  ],
  "students": [
    1
  ],
  "teachers": [
    393
  ]

}
```


api/processors/actions/group/create.py
```diff
@@ -0,0 +1,40 @@
...
+    def _set_group_status(self) -> None:
+        StatusSetter.set_status(
+            obj=self.group, status=GroupStatus.PENDING, status_since=self.timestamp
+        )
+
+    def _set_coordinators_status(self) -> None:
+        return super()._set_coordinators_status()
+
+    def _set_teachers_status(self) -> None:
+        self.group.teachers.update(status=TeacherStatus.GROUP_OFFERED, status_since=self.timestamp)
```
Auch-Auch, line 37: Here I am not sure about business logic , according to #123 this should be correct but according to @ivanBarinsky's code of group abort/start we actually need to update statuses of all teachers (not only ones assigned to this group). @lemontree210 What do you think?

lemontree210: When we start/abort/finish the group, we actually _need_ to only update statuses of teachers (and, in the case of those actions, coordinators too) that were assigned to that group. But that would have led to a need to create more filters IIRC, so we discussed that with @Roxe322 and decided that we'd just update statuses of all teachers (and coordinators) because the statuses depend on number of groups and the result would be the same anyway. Since we use `.count()` to count groups for setting those statuses, no additional DB hits will be required anyway (unless I'm very wrong here).

Another thing that is special about this particular situation (with group being offered) is that, no matter whether the teacher is already working with another group or not, I want this situation (that the teacher was offered a group and either the bot or the coordinator are awaiting their reply ASAP) to be very clear and easily displayable to the coordinators.

The next step, then (this is not pertinent to this PR, just explaining the logic), will depend on whether the teacher and the sufficient number of students confirm their willingness to study: in case of success, we confirm the group and mark it as ready for start (#124) or discard it and reset the statuses for teachers according to how many groups they're still working with (#175).

api/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 4.2.2 on 2023-06-16 15:39
+# Generated by Django 4.2 on 2023-07-09 12:30
```
lemontree210, line 1: I think you forgot to run `poetry install` :)

api/views/group.py
```diff
@@ -9,7 +11,25 @@
...
+    GroupReadSerializer,
+    GroupWriteSerializer,
+    PublicGroupSerializer,
+    PublicGroupWithStudentsSerializer,
+)
+from api.views.mixins import ReadWriteSerializersMixin
+
+
+class GroupViewSet(ReadWriteSerializersMixin, viewsets.ModelViewSet[Group]):  # type: ignore
+    queryset = Group.objects.all()
```
lemontree210, line 24:
```suggestion
class GroupViewSet(ReadWriteSerializersMixin, viewsets.ModelViewSet[Group]):  # type: ignore
    """Viewset for groups. To be used by both coordinators in dashboard (public API) and by bot."""

    queryset = Group.objects.all()
```

lemontree210 (changes requested):
I'm happy with the changes, especially the tests. Thank you!
I would just add one docstring.
Let's see if @Roxe322 has something to say.

api/migrations/0001_initial.py
Auch-Auch: I did, but at some point downgraded to check something, right now I have 4.2.2 but I did not create a new commit because I think it won't brake anything and correct version will be back at next commit.

lemontree210: nah, I was just being fastidious :) This is obviously not a major version change.

lemontree210 (approved):
LGTM. @Roxe322 said he would look at it tomorrow, that much we can wait :)

lemontree210:
@Auch-Auch It's not draft anymore, is it? If you're potentially ready to merge, please convert this PR into "Ready for review" :) I won't do it myself, maybe you're planning to do something else.

tests/tests_api/test_group_internal.py
```diff
@@ -0,0 +1,52 @@
...
+            "teachers": [t.pk for t in group.teachers.iterator()],
+        }
+
+    def test_group_create_general(self, api_client, group: Group, timestamp):
+        data = self.get_data(group)
+        response = api_client.post("/api/groups/", data)
+        assert response.status_code == status.HTTP_201_CREATED
+        created_group = Group.objects.get(id=response.data["id"])
+        serializer = GroupWriteSerializer(created_group)
+        for k, v in data.items():
```
Roxe322, line 31: Please, let's use more readable variable names here

api/urls.py
```diff
@@ -20,6 +21,7 @@
 )
 
 internal_router = DefaultRouter()
+internal_router.register(r"groups", GroupViewSet, basename="groups")
```
Roxe322, line 24: Are you sure it should be internal? As I see, it will be used by Dashboard. Then you probably should either edit PublicGroupViewSet somehow, or create a new one, but in `/public/` zone

Roxe322: Because now "internal" means "not dashboard" (bot or blank DRF/curl interface)

Roxe322 (changes requested):

api/urls.py
Roxe322: We definitely have to discuss permissions, and we even have TODOs about it :)

And also, I understand that our routers naming decision may be confusing, but this is what we decided

Roxe322: And yes, it's fine to either do create route, or use default DRF create method (POST to /public/groups/) for this

Auch-Auch: Added POST /public/groups for group creation, please, have a look one more time

Roxe322 (approved):

api/urls.py
Roxe322: LGTM

lemontree210:
@Roxe322 @Auch-Auch Thanks guys! Merging.

tests/tests_api/test_group.py
```diff
@@ -406,3 +366,48 @@ def test_abort_appends_former_entity_lists(self, api_client, active_group, avail
...
+        assert group.status == GroupStatus.PENDING
+
+        common_status_since = created_group.status_since
+        compare_date_time_with_timestamp(common_status_since, timestamp)
+
+        for student in group.students.iterator():
+            assert student.status == StudentStatus.GROUP_OFFERED
+            assert student.status_since == common_status_since
+            log_event: StudentLogEvent = StudentLogEvent.objects.get(student_id=student.pk)
+            assert log_event.type == StudentLogEventType.GROUP_OFFERED
```
lemontree210, line 405: @Auch-Auch, after discussion in #191 ([comment](https://github.com/samanthasgroup/django-webapps/pull/191#discussion_r1271420281)) I guess that these lines are where `group_to` should have been checked and the test should have failed because LogEventCreator was creating student and teacher events incorrectly.

Same for the `for teacher ...` loop below.

PR #181, opened 2023-07-14
feat(API views): look up bot chat ID, return Chatwoot conversation ID

### LINKED ISSUE #180: Create API view for Bot: get `chatwoot_conversation_id` by `registration_telegram_bot_chat_id`
As we are seeking to connect Chatwoot with [our bot](https://github.com/samanthasgroup/samanthas-telegram-bot), we must be able to pass a message from user via bot to Chatwoot and vice versa.

This architecture implies that one chat ID in Telegram bot corresponds to exactly one conversation ID in Chatwoot. 

I don't think this must be reflected in our models (e.g. by creating a one-to-one relationship), because that would mean that the backend knows something about the connection between Chatwoot and bot, which will be too much knowledge. 

But we do need an API view that gets telegram chat ID and either returns existing Chatwoot conversation ID (if this chat ID is already in the database) or ~status code 406~ empty data, in which case our bot contacts Chatwoot and performs necessary operations ([create contact, create conversation](https://www.chatwoot.com/docs/product/channels/api/send-messages/)) and obtains conversation ID, which it will then pass along with other data to the backend when creating a `PersonalInfo` record.

~This view will then be similar to `check_existence_of_chat_id()` but will return chatwoot conversation ID instead of Telegram chat ID.~ A generic implementation of `PersonalInfo` viewset can be used (with addition of a `FilterSet`), where bot just passes chat_id as URL parameter.

Another solution would be for the backend to contact Chatwoot directly and send conversation ID to the bot. but the downside would be that we would have to implement a Chatwoot client twice, and also that backend will care about how Chatwoot works, which shouldn't be the case.

lemontree210:
closes #180 

**UPD**: After [this comment](https://github.com/samanthasgroup/django-webapps/pull/181#issuecomment-1636294209) I reworked this to just add a URL parameter to the generic Personal Info viewset. The client (bot) will have to extract Chatwoot conversation ID on its own.

**UPD 2**: After a lot of discussion and a [vote](https://t.me/c/1808422267/1258) among the dev team, I removed status code 406 for empty results. Any valid request is responded to with status code 200 and empty list if there are no results.


Auch-Auch (approved):

api/serializers/personal_info.py
```diff
@@ -53,6 +55,36 @@ class Meta:
         extra_kwargs = {"registration_telegram_bot_chat_id": {"required": True}}
 
 
+class GetChatwootConversationIdSerializer(serializers.ModelSerializer[PersonalInfo]):
+    """A serializer used to look up Chatwoot conversation ID by Telegram bot chat ID."""
+
+    chatwoot_conversation_id = serializers.SerializerMethodField(read_only=True)
```
Roxe322, line 61: MethodField is always read-only, by design

Roxe322 (changes requested):

Roxe322:
I think it's better to remove this new action and make a filter for /api/personal_info/ by registration_telegram_bot_chat_id

so the request will look like /api/personal_info/?registration_telegram_bot_chat_id=12345 , then on client side you get first element and its chatwoot_conversation_id

api/serializers/personal_info.py
lemontree210: Well, I made it doubly sure, didn't I? 😆

lemontree210:
> I think it's better to remove this new action and make a filter for /api/personal_info/ by registration_telegram_bot_chat_id
> 
> so the request will look like /api/personal_info/?registration_telegram_bot_chat_id=12345 , then on client side you get first element and its chatwoot_conversation_id

OK, that will definitely make this view more universal. Will look into it.

api/serializers/personal_info.py
lemontree210: This is not relevant as it got removed from the code

lemontree210:
> I think it's better to remove this new action and make a filter for /api/personal_info/ by registration_telegram_bot_chat_id
> 
> so the request will look like /api/personal_info/?registration_telegram_bot_chat_id=12345 , then on client side you get first element and its chatwoot_conversation_id

Done

api/views/personal_info.py
```diff
@@ -14,8 +16,41 @@
...
+    },
+)
 class PersonalInfoViewSet(viewsets.ModelViewSet[PersonalInfo]):
-    queryset = PersonalInfo.objects.all()
+    def get_queryset(self) -> QuerySet[PersonalInfo]:
+        """Optionally restricts the returned personal info items
+        to items with given chat ID in Telegram registration bot.
+        """
+        queryset: QuerySet[PersonalInfo] = PersonalInfo.objects.all()
+        chat_id = self.request.query_params.get("registration_telegram_bot_chat_id")
```
Roxe322, line 44: Let's use django-filters please :) 

Like here:
https://github.com/samanthasgroup/django-webapps/blob/a0f9df55f749140883bc8836741daa6c5b67f3e8/api/filters/age_range.py#L6

Then you won't need to redefine get_queryset at all

Roxe322 (changes requested):

api/views/personal_info.py
lemontree210: Done

Roxe322 (changes requested):

api/views/personal_info.py
```diff
@@ -76,3 +82,30 @@ def check_existence_of_chat_id(self, request: Request) -> Response:
...
+                response=PersonalInfoSerializer,
+                description="Users with these parameters exist",
+            ),
+            status.HTTP_406_NOT_ACCEPTABLE: OpenApiResponse(
+                response=BaseAPIExceptionSerializer,
+                description="No user with these parameters exists",
+            ),
+        }
+    )
+    def list(
```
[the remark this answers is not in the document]
Auch-Auch: I would keep an empty list because, on the user side, you need to create an explicit error handler for 406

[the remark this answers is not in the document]
Auch-Auch: Btw, maybe it is already working that way but, in general, middleware is a good idea, it is usually a good practice to handle all your exceptions (for all codes) in a single place and pass whatever parameter you need in the Exception itself

api/views/personal_info.py
```diff
@@ -76,3 +82,20 @@ def check_existence_of_chat_id(self, request: Request) -> Response:
...
         serializer.is_valid(raise_exception=True)
         return Response(request.query_params)
+
+    @extend_schema(
+        responses={
+            status.HTTP_200_OK: OpenApiResponse(
+                response=PersonalInfoSerializer,
+                description="Users with these parameters exist",
+            ),
+            status.HTTP_406_NOT_ACCEPTABLE: OpenApiResponse(
```
Roxe322, line 92: I suggest to move it to custom schema, because now every GET handler could return 406.

Roxe322 (changes requested):

api/views/personal_info.py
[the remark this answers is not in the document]
lemontree210: > I would keep an empty list because, on the user side, you need to create an explicit error handler for 406

Keeping an empty list as voted by the dev team

api/middleware/custom_status_code.py
```diff
@@ -0,0 +1,31 @@
...
+        self.get_response = get_response
+
+    def __call__(self, request: HttpRequest) -> Response | JsonResponse:
+        response: Response = self.get_response(request)
+
+        if request.method != "GET":
+            return response
+
+        if not response.data:
+            return JsonResponse(
```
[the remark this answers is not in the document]
lemontree210: not relevant: I removed middleware

api/views/personal_info.py
lemontree210: not relevant anymore: returning 200 OK and empty list

Roxe322 (approved):

PR #186, opened 2023-07-15
Renaming of public views/serializers to dashboard ones

### LINKED ISSUE #177: [DISCUSSION] Naming of serializers and viewsets
We have two clients for our views:
- Coordinator's dashboard (ToolJet)
- Bot

We also have views that can be used both by coordinator and by bot (e.g. #174).

In the future we cannot exclude creation of a teacher's dashboard (not in MVP) and admin's dashboard (right now we plan to implement basic functionality for the admins in Django admin site) and views for them. Unfortunately, the free version of ToolJet doesn't support different roles, so right now Tooljet will only provide a dashboard for coordinators. 

I know that the dashboard category was initially called "public", but this was no final decision: we decided that we would take it up later when need arises. 

The objections against "public" are:

- In a way, none of the views are really public because the system will be closed for the outside world
- If more dashboards get created, "public" gets even more confusing.

So I suggest that, without resorting to "brand" names like "ToolJet", we can be sure that we have some bot (or bots) and some dashboard that access data from the backend. So the serializers could have names like `BotTeacherSerializer`, `CoordinatorDashboardTeacherSerializer` (same for viewsets). For the cases when both clients can use the same serializer/viewset, we can create some `CommonTeacherSerializer` or just `TeacherSerializer`.

Auch-Auch:
closes #177
Refactoring of names based on discussion in #177.


api/views/group.py
```diff
@@ -13,15 +13,31 @@
...
+    GroupReadSerializer,
     GroupWriteSerializer,
-    PublicGroupSerializer,
-    PublicGroupWithStudentsSerializer,
 )
+from api.views.mixins import ReadWriteSerializersMixin
 
 
-class PublicGroupViewSet(viewsets.ReadOnlyModelViewSet[Group], CreateModelMixin):
+class GroupViewSet(ReadWriteSerializersMixin, viewsets.ModelViewSet[Group]):  # type: ignore
```
Auch-Auch, line 24: Added this `ViewSet` for API to be consistent, e.g. this view set will be used for internal needs whereas dashboard will be used only for dashboard.

lemontree210:
@Auch-Auch I think we should also create the module `teacher/internal.py` and move some serializers there from `dashboard.py`

lemontree210:
@Auch-Auch note that I merged `master` into your branch after dependabot's updates

api/views/group.py
lemontree210: How about a `CommonGroupViewSet` with `create()` method?
And please add a short docstring :)

api/urls.py
```diff
@@ -39,23 +41,23 @@
...
+    DashboardTeacherWithPersonalInfoViewSet,
     basename="teachers_with_personal_info",
 )
 
-# "Internal" API is used by bot and "public" API is used by the Tooljet
+# "Internal" API is used by bot and "dashboard" API is used by the Tooljet
 urlpatterns = [
     path("", include(internal_router.urls)),
-    path("public/", include(public_router.urls)),
+    path("dashboard/", include(dashboard_router.urls)),
```
lemontree210, line 62: We might want to leave the URL path name "public/" as it is if it would require too much effort from @ivanBarinsky to change URLs in the dashboard. Let's see what Ivan says :)
(Same for other cases)

api/views/group.py
Auch-Auch: Yep, but I think something like `GroupCreationMixin` would be better here, it kind of follow patter of DRF Mixins, added it to the pull request

api/urls.py
ivanBarinsky: It wouldn't cause too many issues on my side, i think its a good time to do it now since im not using too many queries in tooljet quite yet 👍

ivanBarinsky: Happy to resolve this on my side

lemontree210: > It wouldn't cause too many issues on my side, i think its a good time to do it now since im not using too many queries in tooljet quite yet +1

OK, let's do `dashboard/` then :)

lemontree210 (approved):
LGTM. Actually, the bot isn't using any `GroupViewSet` and `groups/` route as of now, but it surely will.

Let's see what @Roxe322 has to say :)

api/views/group.py
```diff
@@ -13,15 +13,39 @@
...
+    GroupReadSerializer,
     GroupWriteSerializer,
-    PublicGroupSerializer,
-    PublicGroupWithStudentsSerializer,
 )
+from api.views.mixins import ReadWriteSerializersMixin
 
 
-class PublicGroupViewSet(viewsets.ReadOnlyModelViewSet[Group], CreateModelMixin):
+class CreateGroupMixin(CreateModelMixin):
```
Roxe322, line 24: I think it's better to put it to either mixins.py or mixins/group.py, since it's not a view

Roxe322 (approved):

Roxe322 (changes requested):
missclicked on approve :)

api/views/group.py
Auch-Auch: Created a separate package for mixins

Roxe322 (approved):

lemontree210:
Thank you for your effort!

Issue #167, opened 2023-07-07
Refactor tests


lemontree210:
Refactoring may include the following steps, as suggested by @Auch-Auch :
- Create a separate package for fixtures, so that for every module in `tests_api` we have a module for its fixtures and place all fake models there. By doing so we decouple bakery and test logic which is preferable in this case e.g. due to uniqueness constraints.
- Make use of serializers to get rid of hardcoded dicts

lemontree210:
closing as fixed by #200

Issue #116, opened 2023-06-11
[COMPOUND] Create API views for actions in ToolJet and bot


lemontree210:
This issue follows up on #30 , introducing actions with coordinators, groups, students, adult and young teachers.

## Groups
- [x] #276
- [x] #123
- [x] #124
- [x] #175
- [x] #117
- [x] #125
- [x] #234

## Students
- [x] #239
In MVP we might not use `REQUESTED_TRANSFER` and the users won't be able to request transfer via bot, there will be no intermediate stage. The coordinator just chooses the new group, and the student (or teacher below) is instantly transferred to the new group.
- [x] #263
- [x] #267
- [x] #275
- [x] #292
- [x] #298 
- [x] #299
- [x] #306
- [x] #312
- [x] #320

\* When the group has finished studies (see #234), the coordinator has to contact the admins and/or the student to find out whether they can get another group. 

## Adult teachers

Note that some API views are the same for students and adult teachers

- [x] #251
See the note above on students
- [x] #267
- [x] #275 
- [x] ~API view: determine next project status after group has finished studies~ _(I think I copied it here from `Student` by mistake)_
- [x] #306 
- [x] #312
- [x] #333
- [x] #334

## Young teachers

In [MVP](https://github.com/samanthasgroup/django-webapps/milestone/1), young teachers will maybe teach speaking clubs "informally" (e.g. not controlled by dashboard). We will have to discuss this.

## Coordinators

In [MVP](https://github.com/samanthasgroup/django-webapps/milestone/1), coordinators cannot perform any actions with other coordinators, so no dashboard views are needed. However, since it was decided that coordinators will register through our bot, one view is needed:

- [x] #283

lemontree210:
## A note on `from_group` and `to_group` in `LogEvent`s created by action processors

While reviewing #191, I realized that I probably didn't have a clear enough vision of how exactly we should fill `from_group` and `to_group` fields in `LogEvent`. So I'll try and formulate it this way:

- If the subject of a `LogEvent` was not in a given `Group` before the event described and became a member of this `Group` as a result of the event, this `Group` is passed into `to_group` parameter at `LogEvent` creation;
- conversely, If the subject of a `LogEvent` was in a given `Group` before the event described and is no longer member of this `Group` as a result of the event, this `Group` is passed into `from_group` parameter at `LogEvent` creation.

This means that:

- If a student is transferred between groups, both `from_group` and `to_group` are filled
[...]

PR #191, opened 2023-07-22
API view: Confirm that `Group` is ready to start

### LINKED ISSUE #124: API view: confirm the group is ready to start studies

Roxe322:
Closes #124

### DIFF
--- a/api/processors/actions/group/abort.py
+++ b/api/processors/actions/group/abort.py
@@ -27,6 +27,7 @@ class GroupAbortProcessor(GroupActionProcessor):
             teacher_log_event_type=TeacherLogEventType.GROUP_ABORTED,
             coordinator_log_event_type=CoordinatorLogEventType.GROUP_ABORTED,
             group_log_event_type=GroupLogEventType.ABORTED,
+            from_group=self.group,
         )
 
     def _move_related_people_to_former(self) -> None:
--- a/api/processors/actions/group/confirm_ready_to_start.py
+++ b/api/processors/actions/group/confirm_ready_to_start.py
@@ -7,7 +7,7 @@ from api.models.choices.log_event_type import (
 )
 from api.models.choices.status import GroupStatus, StudentStatus, TeacherStatus
 from api.processors.actions.group import GroupActionProcessor
-from api.processors.auxil.log_event_creator import FromOrToGroup, GroupLogEventCreator
+from api.processors.auxil.log_event_creator import GroupLogEventCreator
 
 
 class GroupConfirmReadyToStartProcessor(GroupActionProcessor):
@@ -18,7 +18,7 @@ class GroupConfirmReadyToStartProcessor(GroupActionProcessor):
             teacher_log_event_type=TeacherLogEventType.GROUP_CONFIRMED,
             coordinator_log_event_type=CoordinatorLogEventType.TOOK_NEW_GROUP,
             group_log_event_type=GroupLogEventType.CONFIRMED,
-            from_or_to_group=FromOrToGroup.TO,
+            to_group=self.group,
         )
 
     def _set_coordinators_status(self) -> None:
--- a/api/processors/actions/group/confirm_ready_to_start.py
+++ b/api/processors/actions/group/confirm_ready_to_start.py
@@ -18,7 +18,6 @@ class GroupConfirmReadyToStartProcessor(GroupActionProcessor):
             teacher_log_event_type=TeacherLogEventType.GROUP_CONFIRMED,
             coordinator_log_event_type=CoordinatorLogEventType.TOOK_NEW_GROUP,
             group_log_event_type=GroupLogEventType.CONFIRMED,
-            to_group=self.group,
         )
 
     def _set_coordinators_status(self) -> None:
--- a/api/processors/actions/group/confirm_ready_to_start.py
+++ b/api/processors/actions/group/confirm_ready_to_start.py
@@ -1,0 +1,42 @@
+from api.models.auxil.status_setter import StatusSetter
+from api.models.choices.log_event_type import (
+    CoordinatorLogEventType,
+    GroupLogEventType,
+    StudentLogEventType,
+    TeacherLogEventType,
+)
+from api.models.choices.status import GroupStatus, StudentStatus, TeacherStatus
+from api.processors.actions.group import GroupActionProcessor
+from api.processors.auxil.log_event_creator import FromOrToGroup, GroupLogEventCreator
+
+
+class GroupConfirmReadyToStartProcessor(GroupActionProcessor):
+    def _create_log_events(self) -> None:
+        GroupLogEventCreator.create(
+            group=self.group,
+            student_log_event_type=StudentLogEventType.GROUP_CONFIRMED,
+            teacher_log_event_type=TeacherLogEventType.GROUP_CONFIRMED,
+            coordinator_log_event_type=CoordinatorLogEventType.TOOK_NEW_GROUP,
+            group_log_event_type=GroupLogEventType.CONFIRMED,
+            from_or_to_group=FromOrToGroup.TO,
+        )
+
+    def _set_coordinators_status(self) -> None:
+        # status will change when the group starts studies
+        pass
+
+    def _set_group_status(self) -> None:
+        StatusSetter.set_status(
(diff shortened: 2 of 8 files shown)


api/processors/auxil/log_event_creator.py
```diff
@@ -7,14 +9,20 @@
 )
 
 
+class FromOrToGroup(enum.Enum):
```
Auch-Auch, line 12: Enums are kind of broken in Python, I would inherit from str as well, e.g. `class FromOrToGroup(str, enum.Enum)` or add a `__str__` magic method otherwise it might lead to bugs, for example, during serialization (though it is not the case right now).

api/processors/auxil/log_event_creator.py
```diff
@@ -7,14 +9,20 @@
...
+    FROM = "from_group"
+    TO = "to_group"
```
lemontree210, line 14: There is also a possibility that a student or a teacher is transferred from one group to another. Does the general design of your solution support this case?

Also, I think I understand why it's in this module (as it's only related to group events), maybe that's why you didn't declare this class somewhere in e.g. `enums.py`. Just checking.

api/processors/auxil/log_event_creator.py
```diff
@@ -7,14 +9,20 @@
...
 class GroupLogEventCreator:
     @staticmethod
-    def create(
+    def create(  # noqa: PLR0913
         group: Group,
         group_log_event_type: GroupLogEventType,
         student_log_event_type: StudentLogEventType,
         teacher_log_event_type: TeacherLogEventType,
         coordinator_log_event_type: CoordinatorLogEventType | None = None,
+        from_or_to_group: FromOrToGroup = FromOrToGroup.FROM,
```
lemontree210, line 25: Is the default value justified by the fact that the most views we have already done create log events with `group_from` rather than `group_to`? If so, I think that it should be marked as a temporary solution with a TODO and then a separate PR should be prepared where this default value is removed and all views set this arg explicitly. There is no general logic behind this default value, so leaving it like this could make the code confusing.

Also, when a group is created (#174), the log event should have a `group_to`, not `group_from` attribute set. I guess this is a mistake I didn't spot then.

Auch-Auch (approved):

api/processors/auxil/log_event_creator.py
```diff
@@ -26,13 +34,14 @@ def create(
...
 
         StudentLogEvent.objects.bulk_create(
-            StudentLogEvent(student=student, from_group=group, type=student_log_event_type)
+            StudentLogEvent(student=student, type=student_log_event_type, **from_or_to_group_arg)
             for student in group.students.iterator()
         )
 
         TeacherLogEvent.objects.bulk_create(
-            TeacherLogEvent(teacher=teacher, from_group=group, type=teacher_log_event_type)
+            TeacherLogEvent(teacher=teacher, type=teacher_log_event_type, **from_or_to_group_arg)
```
lemontree210, line 45: Hm... I guess it was a mistake in one of the previous PRs, there shouldn't be just `from_group`, of course, there should be both.

lemontree210 (changes requested):
I'm not sure this solution with a `group_to_or_from` will work. The main reason is that there will be log events that will have both `group_from` and `group_to` filled. 

Also, I think I just realized that `GroupLogEventCreator.create()` wrongly specified only `from_group=` when in fact both should be specified. I wonder how it wasn't spotted by the tests in #174 because group creation log events for teacher and student are supposed to have the given group be set as `group_to`, not `group_from`. @Auch-Auch could you look into that?

api/processors/auxil/log_event_creator.py
Roxe322: Yeah, I know about it, but now it's used only as a function param, so I decided that we don't need proper str representation for it.

api/processors/auxil/log_event_creator.py
Roxe322: Redesigned it to use both from_group and to_group, so removed the enum

api/processors/auxil/log_event_creator.py
Roxe322: Now it supports both. And also fixed issue with group creation.

api/processors/auxil/log_event_creator.py
Roxe322: Fixed

api/processors/auxil/log_event_creator.py
lemontree210: enum removed, so hiding this conversation

api/processors/actions/group/start.py
```diff
@@ -19,6 +19,7 @@ def _create_log_events(self) -> None:
             teacher_log_event_type=TeacherLogEventType.STUDY_START,
             coordinator_log_event_type=CoordinatorLogEventType.TOOK_NEW_GROUP,
             group_log_event_type=GroupLogEventType.STARTED,
+            from_group=self.group,
```
lemontree210, line 22:
```suggestion
```
Same as above

api/processors/actions/group/confirm_ready_to_start.py
```diff
@@ -18,7 +18,7 @@ def _create_log_events(self) -> None:
             teacher_log_event_type=TeacherLogEventType.GROUP_CONFIRMED,
             coordinator_log_event_type=CoordinatorLogEventType.TOOK_NEW_GROUP,
             group_log_event_type=GroupLogEventType.CONFIRMED,
-            from_or_to_group=FromOrToGroup.TO,
+            to_group=self.group,
```
lemontree210, line 21:
```suggestion
```
Let's remove this altogether, as per [this explanation](https://github.com/samanthasgroup/django-webapps/issues/116#issuecomment-1646980857)

api/processors/actions/group/create.py
```diff
@@ -23,6 +23,7 @@ def _create_log_events(self) -> None:
             student_log_event_type=StudentLogEventType.GROUP_OFFERED,
             teacher_log_event_type=TeacherLogEventType.GROUP_OFFERED,
             group_log_event_type=GroupLogEventType.FORMED,
+            to_group=self.group,
```
lemontree210, line 26: Here, `to_group` stays as you added it 👍🏻

api/processors/actions/group/abort.py
```diff
@@ -27,6 +27,7 @@ def _create_log_events(self) -> None:
             teacher_log_event_type=TeacherLogEventType.GROUP_ABORTED,
             coordinator_log_event_type=CoordinatorLogEventType.GROUP_ABORTED,
             group_log_event_type=GroupLogEventType.ABORTED,
+            from_group=self.group,
```
lemontree210, line 30: 👍🏻

lemontree210 (changes requested):
I know it's a bit confusing, not least because I didn't form a clear enough concept of `from_group` and `to_group`. I hope it will be clearer to both me and y'all now :)

api/processors/actions/group/start.py
Roxe322: Fixed

api/processors/actions/group/confirm_ready_to_start.py
Roxe322: Fixed

lemontree210 (approved):
I'm approving this one, but we have to keep in mind (@Auch-Auch) that tests might need to be adjusted to catch those `from_group` and `to_group` quirks (in a separate PR, most likely in #167). 

It has to be said that this is definitely not the most important part of our system: the log events will be just for coordinator's information and for potential analytics later on. Nothing will break if something is displayed slightly incorrectly. But the tests will require only a small adjustment IISC.

Issue #102, opened 2023-05-30
[COMPOUND] Refactor tests, add missing tests for API views


Roxe322:
Every API endpoint and method should be covered by test. 

## Refactor tests

Refactoring suggested by @Auch-Auch . 

- [ ] #167

## Add missing tests
- [ ] new APIs `/api/public/students_with_personal_info/` and `/api/public/teachers_with_personal_info/` (created in #100)

PR #200, opened 2023-07-29
Tests refactoring

Auch-Auch:
Link to #102.


lemontree210 (commented):
Thank you for the PR! Our general concern would be that in some cases you removed "literal" data and replaced the explicitly specified fields with serializers' data. This means that you're no longer testing that the API view returns correct data (correct fields), you're just testing whether the correct serializer was used. So should a mistake be made in a serializer, the test will not catch it.

These changes then look like changing the nature of the tests, not like refactoring.

Auch-Auch:
> Thank you for the PR! Our general concern would be that in some cases you removed "literal" data and replaced the explicitly specified fields with serializers' data. This means that you're no longer testing that the API view returns correct data (correct fields), you're just testing whether the correct serializer was used. So should a mistake be made in a serializer, the test will not catch it.
> 
> These changes then look like changing the nature of the tests, not like refactoring.

What do you mean by correct data? API users expects to see data from Serializers, because of the OpenAPI schemas. In terms of user, serializer's data is the only correct one he expects. I mean, what the point in the serializers in the first place.

lemontree210:
By "correct data" I mean that the correct set of fields is returned (and all values that mustn't be empty are not empty).

Here's how I see it: in the "old" version we have sort of integration tests: they test both the routing and the data being produced by the serializers. If you change the tests the way you do, then, as far as I understand, the tests become proper unit tests that only test views, and that is probably a good thing - as long as you then create another set of tests that test only serializers. Which is, maybe, too much of an effort.

Auch-Auch:
I still do not think that this is a big problem, on the contrary each time you add it you also add a circular dependency which might not be too bad now but as API grows it will be worse. But if you so afraid of missing fields we can add a simple check: compare set of fields in response with hardcoded set of fields, but as @lemontree210 correctly mentioned it should be done in a separate layer.

lemontree210:
OK, here's what I suggest that you do (so we don't get stuck):
- create a separate package for serializers' tests
- for those API views' tests that have hard-coded data - create tests for serializers using the same data. Otherwise you'll just delete them and then we'll have to re-add them, which is crazy
- for other API views, follow your own approach and then we'll add tests for serializers with hard-coded data, possibly in a separate PR. It will be either you as well, or someone else who has time

Auch-Auch:
Writing tests for serializers ended up somehow tricky, for some serializers results are different from what we have in the endpoint which is somehow confusing, I did not investigate that deep but seems like some magic is going on in the view, so I decided to revert asserts for previous tests and keep new tests as is.

Auch-Auch:
Should we merge or something else is needed?

lemontree210:
> Should we merge or something else is needed?

I don't know, you didn't re-request review so I'm not sure whether you're done or not 😄

Roxe322 (approved):

lemontree210 (approved):
I trust @Roxe322 TBH so I wasn't thorough enough :) but it does look good

lemontree210:
@Auch-Auch Yay, thank you for the contribution :)

Issue #158, opened 2023-07-03
API views: Status updaters for coordinators and teachers should only affect active members


lemontree210:
Only currently working coordinators and teachers must be affected by status updates. 

For example, at every group start or finish, statuses of coordinators and teachers are updated depending on how many groups they currently have. This operation doesn't make sense for coordinators or teachers that have already left the project or are currently on leave.

lemontree210:
marking as fixed by #202. A separate issue for filtering lists of teachers and students will be created.

PR #202, opened 2023-08-03
Filter active teachers and coordinators in status_setter

Roxe322:
closes (part of?) #158


api/models/choices/status.py
```diff
@@ -33,6 +33,18 @@ class CoordinatorStatus(models.TextChoices):
...
+    @classmethod
+    def active_statuses(cls) -> list["CoordinatorStatus"]:
+        return [
+            cls.PENDING,
+            cls.ONBOARDING,
+            cls.ONBOARDING_STALE,
+            cls.WORKING_BELOW_THRESHOLD,
+            cls.WORKING_OK,
+            cls.WORKING_LIMIT_REACHED,
+            cls.FINISHED_LEFT,
```
ivanBarinsky, line 45: Not sure if thats correct?

ivanBarinsky: I mean just the one on line 45, sounds like if a coordinator has that status, they arent really part of the project anymore

lemontree210, line 45:
```suggestion
```

lemontree210: Yep, I'm adding these notes 👌🏻

api/models/choices/status.py
```diff
@@ -33,6 +33,18 @@ class CoordinatorStatus(models.TextChoices):
     REMOVED = "removed", "All access revoked, accounts closed"
     BANNED = "banned", "Banned from the project"
 
+    @classmethod
+    def active_statuses(cls) -> list["CoordinatorStatus"]:
+        return [
+            cls.PENDING,
+            cls.ONBOARDING,
+            cls.ONBOARDING_STALE,
```
lemontree210, line 41:
```suggestion
```

api/models/choices/status.py
```diff
@@ -97,6 +109,21 @@ class TeacherStatus(models.TextChoices):
     BANNED = "banned", "Banned from the project"
     REMOVED = "removed", "All access revoked, accounts closed"
 
+    @classmethod
+    def active_statuses(cls) -> list["TeacherStatus"]:
+        return [
+            cls.AWAITING_OFFER,
+            cls.GROUP_OFFERED,
+            cls.AWAITING_START,
```
lemontree210, line 117:
```suggestion
```

api/models/choices/status.py
```diff
@@ -97,6 +109,21 @@ class TeacherStatus(models.TextChoices):
...
+            cls.TEACHING_ACCEPTING_MORE,
+            cls.TEACHING_NOT_ACCEPTING_MORE,
+            cls.TEACHING_ANOTHER_GROUP_OFFERED,
+            cls.TEACHING_AWAITING_START_OF_ANOTHER,
+            cls.NEEDS_SUBSTITUTION,
+            # TODO: need to confirm this is correct
+            cls.FINISHED_STAYS,
```
lemontree210, line 124:
```suggestion
```

api/models/choices/status.py
```diff
@@ -113,5 +140,12 @@ class TeacherUnder18Status(models.TextChoices):
     BANNED = "banned", "Banned from the project"
     REMOVED = "removed", "All access revoked, accounts closed"
 
+    @classmethod
+    def active_statuses(cls) -> list["TeacherUnder18Status"]:
+        return [
+            cls.ACTIVE,
+            cls.FINISHED_STAYS,
```
lemontree210, line 147:
```suggestion
```

lemontree210 (commented):
I think that this PR uncovered a flaw in how I put the business logic into the statuses. I think I mixed two types of statuses:
1. permanent status in the project (pending, working, banned, finished_left, finished_stays)
2. temporary status caused by some event (group_offered, awaiting_start, requested_transfer).

Hence those ugly statuses like "working_awaiting_start_of_another". Hence it's hard for me right now to judge which statuses should be deemed "active".

I think that I will need a day or two to figure out the best way of dealing with this. Maybe splitting the statuses will be a good thing because it would better reflect the business logic.

Auch-Auch:
Just a small suggestion from me, usually it is a good idea to simplify statuses at the start and keep only mandatory ones. After we will have real users it will be more obvious what is actually needed.

lemontree210 (approved):

lemontree210:
Thank you!

PR #208, opened 2023-08-06
Split statuses, amend models and views (BREAKING CHANGE)

### LINKED ISSUE #206: [FEATURE/BREAKING]: Split statuses in 2 types and amend existing API views
Right now the statuses for `Coordinator`, `Student` and `Teacher` are mixed in that some of them refer to a permanent state of the respective person in the project (e.g. `WORKING_OK` or `FINISHED_LEFT`) while others are situational (`GROUP_OFFERED`). It became evident especially in #202 that they need to be separated.

- [x] Split statuses, amend models
- [x] Amend view to create `Group`
- [x] Amend view to confirm `Group` is ready to start
- [x] Amend view to start studies in `Group`
- [x] Amend view to abort studies in `Group`

For reference: the original views were created in following PRs:
- #174
- #191
- #119
- #156 

API views for discarding a tentative group and finishing the studies in a group were not created at the time this issue was created, so they don't need to be reworked.

lemontree210:
Closes #206 
@ivanBarinsky Please note that this change will affect the dashboard: `"status"` is getting replaced with `"project_status"`.


api/models/choices/status/project.py
```diff
@@ -0,0 +1,74 @@
...
+    ON_LEAVE = "on_leave", "On leave"
+    LEFT_PREMATURELY = "left_prematurely", "Left the project prematurely"
+    FINISHED = "finished", "Completed the course and left the project"
+    BANNED = "banned", "Banned from the project"
+
+
+class TeacherProjectStatus(models.TextChoices):
+    """Enumeration of possible project-level statuses of a teacher (young or adult)."""
+
+    NOT_WORKING = "not_working", "Not working, waiting for a group"
```
ivanBarinsky, line 59: IMO, if we are already making breaking changes here, maybe we could also change the name of this status along with `NOT_STUDYING` in `StudentProjectStatus` to something that emphasizes the fact that they are waiting for a group instead of just not working/studying? I think it's clear enough now that the PR is fresh in our minds but i could see myself get confused about what they mean as time goes on

ivanBarinsky (commented):
LGTM from a logic standpoint, will leave it up to the django pros to approve it code-wise 🙂 👍 
I imagine that most coordinators will only really care about the newly defined "project status" anyway so thats going to be a nice improvement UI-wise, especially once we move on to server-side filters and start using more friendly filter componetns in the UI. 
Thanks for the PR 👍

api/models/choices/status/project.py
paulin-mipt: I agree. Not working may be perceived as "suspended" as well, while waiting is clearly something we want to action on, especially in the algorithm

api/models/group.py
```diff
@@ -67,10 +67,17 @@ class Group(GroupCommon):
...
-        max_length=DEFAULT_CHOICE_CHAR_FIELD_MAX_LENGTH, choices=GroupStatus.choices
+    project_status = models.CharField(
+        max_length=DEFAULT_CHOICE_CHAR_FIELD_MAX_LENGTH, choices=GroupProjectStatus.choices
+    )
+    situational_status = models.CharField(
+        max_length=DEFAULT_CHOICE_CHAR_FIELD_MAX_LENGTH,
+        choices=GroupSituationalStatus.choices,
+        blank=True,
+    )
+    status_since = models.DateTimeField(
```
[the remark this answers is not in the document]
lemontree210: Of course I have thought about it, but the thing with `status_since` is that it's rather insignificant. It's a mere signal for the coordinator of "when was the last time something, anything, happened to this person". The analytics will be based on log events. And `status_since` can potentially be used to filter out people that have been "stale" for weeks.

So I decided that `situational_status_since` will only confuse the coordinators.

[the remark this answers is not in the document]
Roxe322: Got it

Roxe322 (approved):

api/models/choices/status/project.py
lemontree210: OK, I can replace them with something different. At the same time I want to emphasize with this status that the teacher is not just waiting for some group, they're waiting for the first one.

How about `IDLE`? I mean, I think that every developer would understand what that means: a person is waiting for something to happen, but nothing is happening at the moment :) 

@paulin-mipt @ivanBarinsky what do you think?

lemontree210: UPD my next suggestion is `NO_GROUP_YET`

lemontree210: Renamed into `NO_GROUP_YET`, marking as resolved, merging.

PR #209, opened 2023-08-06
Add initial structure for automatic group creation

### LINKED ISSUE #220: AAGC: Add basic code structure to create a tentative group

paulin-mipt:
Closes #220 

This code explicitly leaves out order of group prioritisation, to avoid making this PR too big. Corresponding method currently yields one "dummy" group instead so we could test code with the script.

Some outstanding questions are left inside the code as comments.


api/group_creation/group_manager.py
```diff
@@ -0,0 +1,150 @@
...
+                    status=next_teacher_status,
+                    status_since=group_creation_timestamp,
+                )
+                GroupManager._create_log_events(group)
+                return group
+        # No suitable groups found
+        return None
+
+    @staticmethod
+    def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
```
paulin-mipt, line 113: Note: I don't like this method, hope to come up with more readable code eventually

api/group_creation/group_manager.py
```diff
@@ -0,0 +1,150 @@
...
+            communication_language_mode=CommunicationLanguageMode(
+                teacher.personal_info.communication_language_mode
+            ),
+            teacher=teacher,
+            students=list(range(1, 10)),  # type: ignore
+            monday=timezone.now(),
+            wednesday=timezone.now(),
+        )
+
+    # QQ: do we need this?
```
lemontree210, line 142: Yes, we do need to create log events. BTW please have a look at `models/log_event.py`. I actually think that it was wrong of me to put `on_delete=models.CASCADE` for `.group` in all subclasses of `LogEvent` except `GroupLogEvent`. This is because we will most likely be deleting tentative groups as I don't see any reason to keep thousands of auto-created tentative groups in the database. The log events, however, I would want to keep, to see how many times a student and a teacher refused to take on a group. It should definitely remain `CASCADE` in GroupLogEvent, though.

api/group_creation/group_manager.py
lemontree210, line 1: how about calling the whole package `algorithms` or `jobs` to be consistent with the project structure? `group_manager.py` itself may move to something like `auxil` subpackage - or even to `processors.auxil`. I definitely wouldn't want to have both `GroupManager` and `GroupActionProcessor` in the same project :)

api/group_creation/group_manager.py
lemontree210: I'm not commenting on this bit, then :)

api/group_creation/group_manager.py
```diff
@@ -0,0 +1,150 @@
...
+            return GroupManager.GroupSizeRestriction(min_number=2, max_number=8)
+        if 0 <= age_range.age_from <= age_range.age_to <= MAX_AGE_KIDS_GROUP:
+            return GroupManager.GroupSizeRestriction(min_number=2, max_number=6)
+        # If age range does not fall into expected ranges, fail early
+        raise ValueError(f"Group age range is inconsistent with boundaries: {age_range}")
+
+    @staticmethod
+    def _iterate_group_candidates(teacher_id: int) -> Iterable[GroupCandidate]:
+        # THIS IS A STUB METHOD RETURNING DUMMY VALUES
+        # TODO filter for language+level, communication language, age groups, time slots (all UTC?)
```
lemontree210, line 126: yes, all time slots are UTC. This was intended to help the automatic algorithm right from the start.

api/group_creation/group_manager.py
```diff
@@ -0,0 +1,150 @@
...
+            return GroupManager.GroupSizeRestriction(min_number=5, max_number=10)
+        if MAX_AGE_KIDS_GROUP <= age_range.age_from <= age_range.age_to <= MAX_AGE_TEEN_GROUP:
+            return GroupManager.GroupSizeRestriction(min_number=2, max_number=8)
+        if 0 <= age_range.age_from <= age_range.age_to <= MAX_AGE_KIDS_GROUP:
+            return GroupManager.GroupSizeRestriction(min_number=2, max_number=6)
+        # If age range does not fall into expected ranges, fail early
+        raise ValueError(f"Group age range is inconsistent with boundaries: {age_range}")
+
+    @staticmethod
+    def _iterate_group_candidates(teacher_id: int) -> Iterable[GroupCandidate]:
```
lemontree210, line 124:
```suggestion
    def _iterate_group_candidates(teacher_id: int) -> Iterator[GroupCandidate]:
```
see SO discussion [here](https://stackoverflow.com/a/57916822).

lemontree210 (changes requested):
Thanks for the PR! 🚀 

Please note that statuses will change after #208 is merged. This will be an unexpected breaking change due to my mistake in building the business logic 😞.

api/group_creation/group_manager.py
```diff
@@ -0,0 +1,150 @@
...
+                    friday=group_candidate.friday,
+                    saturday=group_candidate.saturday,
+                    sunday=group_candidate.sunday,
+                )
+
+                group_creation_timestamp = timezone.now()
+                StatusSetter.set_status(
+                    obj=group, status=GroupStatus.PENDING, status_since=group_creation_timestamp
+                )
+                group.save()
```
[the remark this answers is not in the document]
paulin-mipt: That was actually necessary to add many to many connections for teachers and students further on. Otherwise the db transaction was failing because the group id is not saved yet and it couldn't be referred to as a foreign key.

api/group_creation/group_manager.py
```diff
@@ -0,0 +1,150 @@
...
+                group.teachers.add(group_candidate.teacher)
+                group.students.set(group_candidate.students)
+
+                # QQ: do we update teacher/student status automatically,
+                # or wait for coordinator action?
```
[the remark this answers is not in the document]
paulin-mipt: Ok, great.  What kind of event the bot is triggered on though? "Offered" status sounds to me as if it's already offered, not just "ready to offer". But maybe it's fine, I don't know the intended underlying mechanic.

api/group_creation/group_manager.py
```diff
@@ -0,0 +1,150 @@
...
+from api.models.choices.status import GroupStatus, TeacherStatus
+from api.models.language_and_level import LanguageAndLevel
+from api.processors.auxil.log_event_creator import GroupLogEventCreator
+
+
+class GroupManager:
+    @dataclass
+    class GroupSizeRestriction:
+        min_number: int
+        max_number: int
```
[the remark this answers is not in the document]
paulin-mipt: Probably, yes. I'm usually just wary to use reserved standard library tokens even as attributes, just in case

paulin-mipt:
> Please note that statuses will change after #208 is merged. This will be an unexpected breaking change due to my mistake in building the business logic 😞.

No prob! I'll wait for it to be merged so I'll request review again to make sure that I understand new statuses correctly in regards to the algorithm conditions and status changes

api/group_creation/group_manager.py
paulin-mipt: I'm not sure I follow. So, currently all log events are deleted as CASCADE but it sounds reasonable to me. If we delete group-related log events, won't log events for students and teachers stay in place for analytics still? We wll delete a lot of groups, yes, but we'll keep students and teachers.

api/group_creation/group_manager.py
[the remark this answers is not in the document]
lemontree210: I understand that, but we do use `id` a lot as an attribute, although it is reserved. Anyway, it is of course just a matter of taste :)

api/group_creation/group_manager.py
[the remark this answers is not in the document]
lemontree210: Oh I see.

api/group_creation/group_manager.py
```diff
@@ -0,0 +1,150 @@
...
+            restrictions = GroupManager._get_allowed_group_size(self.age_range)
+            students_num = len(self.students)
+            if restrictions.min_number > students_num:
+                return False
+            if restrictions.max_number < students_num:
+                raise ValueError(
+                    f"Group candidate must respect upper group boundaries, "
+                    f"but {students_num} > {restrictions.max_number}"
+                )
+            return True
```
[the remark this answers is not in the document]
paulin-mipt: That is not what I meant exactly. A suggested group with too few students is a "normal thing", which means "we couldn't find enough students with the higher priority, so we are not creating this group and go further".

If group suggestion code returned more than maximum allowed number of students, we should fail early because it means that preceding logic is wrong. The code sorting students and limiting their number should take it all into account at the same time, so that's just an assertion.

That particular assertion (on max limit) could happen in post-init method instead, that's true. But this sounds like a matter of personal preference to me.

api/group_creation/group_manager.py
paulin-mipt: If you have any specific suggestions, feel free to let it out! :)

api/group_creation/group_manager.py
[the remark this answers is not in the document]
lemontree210: Well the mechanic should be that as soon as the group is formed (by algorithm or coordinator), the back-end sends a `POST` request to a webhook in the bot and it immediately contacts the students and the teacher :) So "offered" should be fine because a coordinator/admin will get a bit more confused by a "ready to offer" instead of "offered" in the log.

api/group_creation/group_manager.py
paulin-mipt: > how about calling the whole package algorithms or jobs to be consistent with the project structure?
Agree, makes sense. Renamed to `algorithms`

> I definitely wouldn't want to have both GroupManager and GroupActionProcessor in the same project :)
Could you elaborate, please?

api/group_creation/group_manager.py
lemontree210: Maybe I misunderstand the mechanic of `CASCADE`, but I think that if you delete a `Group` then, in case of `on_delete=CASCADE`, all log events that were related to this group will also be deleted. And student- and teacher-related log events have `.to_group` and `.from_group` attributes, linking these log events to `Group`. So I'm guessing that these log events will be deleted as well if a tentative `Group` is deleted.

api/group_creation/group_manager.py
lemontree210: > > I definitely wouldn't want to have both GroupManager and GroupActionProcessor in the same project :)
> 
> Could you elaborate, please?

It just sounds so confusing. How, judging by the name, can you tell how exactly `Manager` is different from `Processor`? If one manages groups and the other actions with the groups, why aren't they called `GroupProcessor` and `GroupActionProcessor` and why are they in totally different packages?

The way it is beginning to shape, I think this `Manager` will be a pretty universal auxiliary thing, not necessarily closely related to the inner workings of the algorithm.

api/group_creation/group_manager.py
lemontree210: > If you have any specific suggestions, feel free to let it out! :)

Nah, I was being lazy and skipped the whole bit 😜

api/group_creation/group_manager.py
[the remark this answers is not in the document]
lemontree210: > That is not what I meant exactly. A suggested group with too few students is a "normal thing", which means "we couldn't find enough students with the higher priority, so we are not creating this group and go further".

OK, I think I understand what you mean. But on the other hand, if our algorithm tries to create a group with too few students, how is that fundamentally different from creating a group with too many? Maybe the algorithm shouldn't even attempt to create a group in that case? Or maybe, following the Pythonic fashion of using `try...except...` instead of `if...else...`, maybe the algorithm should try to create a group with too few students or too many and catch the exception?

It's just that when I look "from the outside", I can barely see the reason why the one is normal case, while the other one is a coding error. I do see it after thinking about it for a couple of minutes :)

By the way, when I think of it, it seems that my suggested `.is_acceptable` is completely redundant. What if we scrap that method altogether and just try and create a group and if it's not acceptable, it throws an error in `.post_init`, then the algorithm catches it and goes on without even calling the `.is_acceptable` method?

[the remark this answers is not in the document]
lemontree210: > That particular assertion (on max limit) could happen in post-init method instead, that's true. But this sounds like a matter of personal preference to me.

Yes, definitelu a matter of personal preference. I suggested separating it to keep the distinction between a coding error (that should trigger an exception as early as possible, even before calling any methods) and a meaningful `False`.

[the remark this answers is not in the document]
lemontree210: I guess my general question is - why don't you want to just throw exceptions in any case of limits not being observed? To me, it will make the logic more consice and predictable. But I may be missing something.

scripts/run_group_creation.py
```diff
@@ -0,0 +1,12 @@
...
+
+import django
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "django_webapps.settings")
+django.setup()
+
+from api.algorithms.group_manager import GroupManager  # noqa: E402
+
+if __name__ == "__main__":
+    for teacher in GroupManager.get_available_teachers():
```
Roxe322, line 11: Let's do it as a django management-command? https://docs.djangoproject.com/en/4.2/howto/custom-management-commands/

api/algorithms/group_manager.py
```diff
@@ -0,0 +1,148 @@
...
+            return GroupManager.GroupSizeRestriction(min_number=2, max_number=6)
+        # If age range does not fall into expected ranges, fail early
+        raise ValueError(f"Group age range is inconsistent with boundaries: {age_range}")
+
+    @staticmethod
+    def _iterate_group_candidates(teacher_id: int) -> Iterator[GroupCandidate]:
+        # THIS IS A STUB METHOD RETURNING DUMMY VALUES
+        # TODO filter for language+level, communication language, age groups, time slots (all UTC?)
+        # TODO iterate in correct priority order, yield results
+        teacher = Teacher.objects.filter(personal_info__id=teacher_id).get()
```
Roxe322, line 127: `Teacher.objects.get(personal_info__pk=teacher_id)`

api/algorithms/group_manager.py
```diff
@@ -0,0 +1,148 @@
...
+        sunday: datetime | None = None
+
+        def is_acceptable(self) -> bool:
+            # Checks group size is appropriate for age range
+            restrictions = GroupManager._get_allowed_group_size(self.age_range)
+            students_num = len(self.students)
+            if restrictions.min_number > students_num:
+                return False
+            if restrictions.max_number < students_num:
+                raise ValueError(
```
Roxe322, line 55: Why not return False here? Or why not raise ... in case of `restrictions.min_number > students_num` ?

lemontree210: We were having the same discussion [here](https://github.com/samanthasgroup/django-webapps/pull/209#discussion_r1285262948), so why don't you join us 😄

api/group_creation/group_manager.py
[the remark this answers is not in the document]
Roxe322: I also prefer consistency, and from my point of view - there are no such big difference between any case where limit is violated

api/algorithms/group_manager.py
lemontree210: Hiding this as a duplicate

api/group_creation/group_manager.py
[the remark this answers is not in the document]
paulin-mipt: The logic I had in mind does have have a major difference. The group candidate generated by different database filters can easily return less than minimum people and it's alright, no exception needed. We just don't pick this group while iterating candidates in the order of priority. Filters should be smart enough to put maximum restriction in place before returning the "candidate", so violating this is an inconsistency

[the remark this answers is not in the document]
paulin-mipt: But I can put this extra check inside the main generation logic and not here. Then validation will look different, as you both see fit. I don't want to argue on it, so let's agree and move forward :)

api/group_creation/group_manager.py
paulin-mipt: I agree it could be called something else (GroupBuilderAlgorithm?). Manager is not descriptive and can be confused, this is true. I also don't think it's too bad for the purpose but happy to take any suggestion so we can move forward without sitting on naming.

I think these methods don't belong to GroupProcessor because it does not have the group as a parameter. It's a separate logic (as I see it). My take based on all this will be moving it to `auxil.algorithms`? I may lack context as the project is fairly big, if you disagree - please suggest how you see it exactly.

paulin-mipt: Feel free to unresolve this thread if my suggestion doesn't work for you

api/group_creation/group_manager.py
paulin-mipt: Oh yes! I missed there were references to groups in these events too. Then, would setting to `None` instead work for us? Actually I don't see other ways, but to confirm.

Also I believe we should manually delete group-related log events, so they don't hang around as they are meaningless without the group? I'll do that now

api/group_creation/group_manager.py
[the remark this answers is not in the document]
paulin-mipt: Ah great! Not a question for this PR actually, but what's the recovery mechanism going to be? If the webhook action fails (bot service is down/etc), will we know which ones to retry? That sounds like a critical path.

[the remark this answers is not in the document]
paulin-mipt: That's where "ready to offer" vs "offered" could come handy. The ones that triggered 200 OK from webhook and the ones that didn't for any reason. Just thoughts, we can talk about it someplace else

scripts/run_group_creation.py
paulin-mipt: That is cool, thanks for suggestion! I was suspecting there should've been something better than manual Django init!

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,164 @@
...
+        friday: datetime | None = None
+        saturday: datetime | None = None
+        sunday: datetime | None = None
+
+        def __post_init__(self) -> None:
+            # Validate group size against age-related restrictions
+            restrictions = GroupBuilderAlgorithm._get_allowed_group_size(self.age_range)
+            restrictions.validate_size(len(self.students))
+
+    @staticmethod
```
paulin-mipt, line 71: @lemontree210 can you please advise if this is correct after merging from main?
We used to have "teaching_accepting_more" status, what's the equivalent going to be? Should I just check maximum number of the groups vs existing number? What statuses should we update for such teacher?

lemontree210: > have "teaching_accepting_more" status, what's the equivalent going to be? 
The equivalent will be simply `WORKING`. Those `[NOT]_ACCEPTING_MORE` statuses turned out to be problematic.

> Should I just check maximum number of the groups vs existing number?
Exactly, that's what you should do :)

> What statuses should we update for such teacher?
Their `project_status` does not change until the group starts. The `situational_status` changes to `GROUP_OFFERED` once a tentative group is created, but I think I already did that in #208. So the algorithm just builds a group without changing any statuses, the group creation view should be responsible for that. I don't know if how exactly you're planning to trigger this view, though. Maybe the status-setting functionality from group creation view should be factored out so that both your algorithm and the group creation view can use it.

api/models/log_event.py
```diff
@@ -94,14 +94,14 @@ class StudentLogEvent(LogEvent):
         Group,
         null=True,
         blank=True,
-        on_delete=models.CASCADE,
+        on_delete=models.SET_NULL,
```
lemontree210, line 97: oh, right, I said `PROTECT` earlier, but that was wrong, of course. `SET_NULL` is what we need :)

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,171 @@
...
+)
+from api.models.language_and_level import LanguageAndLevel
+from api.processors.auxil.log_event_creator import GroupLogEventCreator
+
+
+class GroupBuilderAlgorithm:
+    @dataclass
+    class GroupSizeRestriction:
+        # both min and max student numbers are inclusive:
+        # group size must fit into [min; max]
```
lemontree210, line 36: I had a question about the rationale behind making the upper boundary inclusive, but I understand I'll just be wasting everybody's time :) So let's at least make it a docstring so one can see this in an IDE

```suggestion
        """Restriction of a group size. 
        
        Note:
            Both min and max student numbers are inclusive: group size must fit into [``min``; ``max``].
        ```
```

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,171 @@
...
+                    f"but {size} > {self.max}"
+                )
+
+    @dataclass
+    class GroupCandidate:
+        language_and_level: LanguageAndLevel
+        communication_language_mode: CommunicationLanguageMode
+        age_range: AgeRange
+        teacher: Teacher
+        students: list[Student]
```
lemontree210, line 58: maybe a `Sequence`? I mean, at least tuples should be allowed, why not

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,171 @@
...
+        def __post_init__(self) -> None:
+            # Validate group size against age-related restrictions
+            restrictions = GroupBuilderAlgorithm._get_allowed_group_size(self.age_range)
+            restrictions.validate_size(len(self.students))
+
+    @staticmethod
+    def get_available_teachers() -> Iterable[Teacher]:
+        return Teacher.objects.filter(
+            project_status=TeacherProjectStatus.NO_GROUP_YET,
+            situational_status="",
```
lemontree210, line 76: so yeah, unfortunately it gets a bit more complicated here after it got simpler there (i.e. in `project_status`) 😄 We have to include teachers that have `NO_GROUP_YET` and `WORKING` teachers whose `.can_take_more_groups` is `True`. You can check `.can_take_more_groups` for **both** `NO_GROUP_YET` and `WORKING` because we will set `simultaneous_groups` to `0` for teachers that don't want any regular groups and only want to host speaking clubs. So they might have `project_status=NO_GROUP_YET` but they won't be expecting to get any offers for regular groups at all.

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+
+    @staticmethod
+    def is_teacher_available(teacher: Teacher | None) -> bool:
```
lemontree210, line 80:
```suggestion
    def teacher_is_available(teacher: Teacher | None) -> bool:
```

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,171 @@
...
+            teacher.project_status == TeacherProjectStatus.NO_GROUP_YET
+            and teacher.situational_status == ""
+        )
+
+    @staticmethod
+    def create_and_save_group(teacher_id: int) -> Group | None:
+        group_candidate = GroupBuilderAlgorithm._get_group_candidate(teacher_id)
+        if group_candidate is None:
+            # No suitable groups found
+            # TODO: maybe log something?
```
lemontree210, line 93: we can leave it as a TODO. I would think that in this situation you will send something to my bot's webhook, and the bot will notify the admins with its logging and notification functionality.

api/management/commands/create_groups.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def handle(self, *_: str, **options: Any) -> None:
+        teachers: Iterable[Teacher] = []
+        if options["teacher_ids"]:
+            teachers = [
+                Teacher.objects.get(personal_info__id=teacher_id)
+                for teacher_id in options["teacher_ids"]
+            ]
+            for teacher in teachers:
+                if not GroupBuilderAlgorithm.is_teacher_available(teacher):
```
lemontree210, line 27:
```suggestion
                if not GroupBuilderAlgorithm.teacher_is_available(teacher):
```

api/management/commands/create_groups.py
```diff
@@ -0,0 +1,34 @@
...
+                    raise CommandError(f"Teacher is not available: {teacher}")
+        else:
+            # Get all available teachers if none specified
+            teachers = GroupBuilderAlgorithm.get_available_teachers()
+
+        for teacher in teachers:
+            GroupBuilderAlgorithm.create_and_save_group(teacher.personal_info.pk)
```
lemontree210, line 34: I think somewhere here should be a POST request to my bot? If it seems like the correct spot, we could add a TODO here that this is where a webhook call will be made.

lemontree210 (commented):
Thank you! :) I think I'm happy, but unfortunately we do have to wait until we get @Roxe322 's comments, and that will be later on Sunday.

api/processors/auxil/algorithms.py
lemontree210: Another thing: how about creating a property `.is_available` in `Teacher` model? Then you can filter by it and get rid of the `is_teacher_available` static method completely.

api/processors/auxil/algorithms.py
paulin-mipt: Got it, that's clearer now!

Outstanding question: what do we do with teachers taking on more groups? Situational status still changes to "offered"? Is it ok if they actively work on the first group, but not in "working" status anymore, because they are offered another one?

api/processors/auxil/algorithms.py
paulin-mipt: I see, good question though! The only one was (probably...) that I was translating min and max restrictions from our discussions into my head and to the code without changing. Now I'm thinking about it and it's even not bad, even though not conventional (which I wasn't thinking about initially, again).

Now I think it may be more readable if we want to create "groups of 5-10 people", and hence the restriction values are 5 and 10. I think in this particular case it might create less confusion. Or maybe for some people - more confusion. I don't know, haha. Docstring is a proper idea either way!

api/processors/auxil/algorithms.py
paulin-mipt: Fair. And that depends on what django will allow to create manytomany (that's what it's used for). I'll google later...

api/processors/auxil/algorithms.py
paulin-mipt: Thanks for both comments. Sounds very fair, will do

api/processors/auxil/algorithms.py
lemontree210: Well that's why I separated `project_status` from `situational_status`: in the situation you're describing (a teacher is working, say, with 1 group but can take on 2 groups in total, so one more group), the teacher who is offered a new group will keep their `project_status=WORKING` but the `situational_status` will change to `GROUP_OFFERED`. Then, once the communication around that second group results in the group either being confirmed or discarded, the `situational_status` is set to `""` again.

api/processors/auxil/algorithms.py
lemontree210: Well we did the same (inclusive upper bound) for age ranges for the same reason (avoiding confusion for non-programmers), so why not do it here as well. As long as it's well-documented :)

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,171 @@
+# DO NOT USE IN PROD, WORK IN PROGRESS
+
```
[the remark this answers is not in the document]
lemontree210: Thanks, I'll take a look @Roxe322 . I definitely know now that creating `/aux/` modules is a very nasty anti-pattern as it makes code that is run under Windows break 😆

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,171 @@
...
+        return Teacher.objects.filter(
+            project_status=TeacherProjectStatus.NO_GROUP_YET,
+            situational_status="",
+        )
+
+    @staticmethod
+    def is_teacher_available(teacher: Teacher | None) -> bool:
+        if teacher is None:
+            return False
+        return (
```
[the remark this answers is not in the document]
lemontree210: Ah yes, @Roxe322 did this to me too :)

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,171 @@
...
+            sunday=group_candidate.sunday,
+        )
+
+        group_creation_timestamp = timezone.now()
+        StatusSetter.set_status(
+            obj=group,
+            project_status=GroupProjectStatus.PENDING,
+            status_since=group_creation_timestamp,
+        )
+        group.save()
```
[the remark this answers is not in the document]
lemontree210: @Roxe322 I think something similar [was discussed here](https://github.com/samanthasgroup/django-webapps/pull/209#discussion_r1285241254). Don't know if it's really a similar thing.

api/processors/auxil/algorithms.py
paulin-mipt: Actually `can_take_more_groups` is the same as proposed `is_available`, as far as I understand.d

api/processors/auxil/algorithms.py
paulin-mipt: No longer relevant, but out of curiosity - why? It's a common convention to put one of the verbs at the beginning, especially for for methods returning bool, such as is_, if_, has_, contains_...

api/management/commands/create_groups.py
paulin-mipt: If you can provide references for later, this will be great (which particular request?)

api/processors/auxil/algorithms.py
paulin-mipt: What if processes happen to 2+ groups at the same time, though? One group wants to set status to AWAITING_START and the other is being GROUP_OFFERED? Can there be any race conditions, groups getting lost, etc?

api/management/commands/create_groups.py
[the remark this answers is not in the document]
paulin-mipt: That's very convenient, never realised it could work this way.

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,171 @@
...
+    TeacherProjectStatus,
+    TeacherSituationalStatus,
+)
+from api.models.language_and_level import LanguageAndLevel
+from api.processors.auxil.log_event_creator import GroupLogEventCreator
+
+
+class GroupBuilderAlgorithm:
+    @dataclass
+    class GroupSizeRestriction:
```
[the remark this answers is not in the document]
paulin-mipt: If that's a project convention overall, then ok. I personally believe if something is a detail of the class implementation and not intended to be used outside, it shouldn't be exposed on a module level to avoid confusion.

api/processors/auxil/algorithms.py
```diff
@@ -0,0 +1,171 @@
...
+        saturday: datetime | None = None
+        sunday: datetime | None = None
+
+        def __post_init__(self) -> None:
+            # Validate group size against age-related restrictions
+            restrictions = GroupBuilderAlgorithm._get_allowed_group_size(self.age_range)
+            restrictions.validate_size(len(self.students))
+
+    @staticmethod
+    def get_available_teachers() -> Iterable[Teacher]:
```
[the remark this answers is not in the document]
paulin-mipt: Logic changed since the comment, now it's not even a QuerySet :)

Otherwise I feel like sometimes moving one level of abstraction up improves readability, if the actual return type is overly specific.

api/processors/auxil/algorithms.py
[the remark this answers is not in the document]
paulin-mipt: Not relevant anymore. I have questions about using "not" instead of explicit checks though, given potential `None` vs "" issues. Maybe not here, but still.

api/processors/auxil/algorithms.py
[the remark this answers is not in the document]
paulin-mipt: Yes, this is it. Just for the sake of foreign keys validity, it's not avoidable

api/processors/auxil/algorithms.py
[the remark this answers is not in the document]
paulin-mipt: I agree with using sorts of "utils" being an antipattern. At the same time I haven't seen any extensive codebases really being able to avoid it completely. So Ithink we should try avoiding this, but still cut ourselves some slack if it takes too much time.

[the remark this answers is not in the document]
paulin-mipt: > I suggest to do api/processors/services/group_builder.py or even without /services/ and also rename GroupBuilderAlgorithm to GroupBuilderService, or just GroupBuilder.

That's exactly how it was in the first attempt (not in terms of path, but in terms of naming module and class). Before I rename, can @Roxe322 and @lemontree210 agree on this, please?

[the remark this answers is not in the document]
Roxe322: @lemontree210 What do you think about 
`api/processors/services/group_builder.py:GroupBuilder`

api/management/commands/create_groups.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class Command(BaseCommand):
+    """
+    Triggers automatic group creation.
+    If teacher_ids are not specified, triggers for all available teachers.
+    """
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument("--teacher_ids", nargs="*", type=int)
```
Roxe322, line 17: Isn't it more correct to call it `teacherS_ids`? I always do so 🤔 
@lemontree210

api/processors/auxil/algorithms.py
lemontree210: I have thought about it, of course... As far as I understand, right now we're using a synchronous DB engine, so this shouldn't be possible. 

I would also avoid offering two groups to the same teacher at the same time - that could be reflected in an algorithm. I would want the teacher to accept one group and then only after they start teaching, offer another group (if their `simultaneous_groups` is greater than `1`).

This could be done by only adding teacher to a hypothetical group if their `situational_status` is blank (`""`). I guess that even the `TRANSFER_REQUESTED` situational status should preclude a new group from being offered (because a teacher is in a middle of some situation that requires attention).

api/management/commands/create_groups.py
lemontree210: Well I don't know the route yet, but it will be a request that signals the bot that it must send notifications to students and teacher and collect their reactions on whether they are ready to study/teach.

api/processors/auxil/algorithms.py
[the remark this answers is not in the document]
lemontree210: > @lemontree210 What do you think about `api/processors/services/group_builder.py:GroupBuilder`

OK, let's make it a `Builder`. I think it is rather easy to understand the difference between `Builder` and `Processor`.

api/processors/auxil/algorithms.py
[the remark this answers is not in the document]
lemontree210: Yeah, @Roxe322 ruthlessly imposed this convention on us! 😆 Let's keep the classes at module level, then. It's not a big deal, especially when we list the classes to import in `__init__.py` and its `__all__`.

api/management/commands/create_groups.py
lemontree210: Well, according to English grammar, `teacher_ids` is correct in this case. But we don't need to impose this on a project level I don't think 😄

api/processors/auxil/algorithms.py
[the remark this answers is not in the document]
lemontree210: > util

You know what? I just read this article 😆 It actually says that an `utils` **module** is bad, not a package with descriptive names. If we have utils for several packages and put them into separate sub-packages of the respective packages, storing them in modules with descriptive names, then this is not really an anti-pattern, is it?

api/processors/auxil/algorithms.py
paulin-mipt: This makes a lot of sense

api/processors/auxil/algorithms.py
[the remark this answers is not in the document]
paulin-mipt: > api/processors/services/group_builder.py:GroupBuilder

Done!

api/processors/auxil/algorithms.py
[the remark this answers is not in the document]
paulin-mipt: Yes, why not! Did this already

lemontree210 (approved):
LGTM

lemontree210:
@paulin-mipt Thank you for your contribution!

PR #221, opened 2023-08-12
API view: Discard a tentative group that was not confirmed, adjust log event types

### LINKED ISSUE #175: API view: discard the tentative group if the teacher didn't confirm or not enough students confirmed
When a coordinator (from dashboard) or bot send a request stating that a group they created cannot be confirmed because not enough students confirmed their participation or a teacher didn't confirm it.

## Preparation: Delete/add log event types 
- [x] Delete two `NOT_CONFIRMED_...` types of `GroupLogEvent`: we will just discard the tentative group completely, deleting it from database, so no additional log events are actually needed for it.
- [x] At the same time, a teacher/student has to get a log event when they do not decline, but then not enough students confirm their readiness to study.

## 1. Create log events:

Beside individual log events for students and teachers that will be listed in other issues (relating to their individual reactions to an offer), create following `StudentLogEvent`s and `TeacherLogEvent`s of type `TENTATIVE_GROUP_DISCARDED`. As a `.comment`, put a reason why. We can standardize these comments in constants. Right now I can only think of the following options:
- "teacher_declined"
- "teacher_did_not_react"
- "not_enough_students_accepted"

No log event is needed for the coordinator.

## 2. Remove coordinator, students and teacher from the group, then delete the group in question

Actually, I'm not 100% sure what happens if we delete a group. I think that it should automatically disappear from `coordinator/student/teacher.groups`. If it does, there's no need to remove them manually before deleting the group. Since the group was only tentative, there's no need to transfer anything to `..._former` fields.

## 3. Set statuses
- Coordinators' statuses change depending on the number of groups each of them has:
  - if it is below threshold: `WORKING_BELOW_THRESHOLD`
  - if it is above threshold and below the limit: `WORKING_OK`
  - if it has reached the limit: `WORKING_LIMIT_REACHED`
- Its students' status changes to `NO_GROUP_YET` (actually it's theoretically possible for a student to attend two groups in the future (which means that discarding one group wouldn't necessarily mean they are not studying in any group), but it is definitely not planned at this moment and it should be rare)
- Its teachers' statuses change depending on the number of the groups they were ready to teach simultaneously:
  - if it is `0`: `NO_GROUP_YET`
  - if it is more then zero:  `WORKING` 

lemontree210:
closes #175


api/processors/actions/group/discard.py
```diff
@@ -0,0 +1,62 @@
...
+        self._set_statuses()
+        self._create_log_events()
+        self._delete()
+
+    def _create_log_events(self) -> None:
+        GroupLogEventCreator.create(
+            group=self.group,
+            student_log_event_type=StudentLogEventType.TENTATIVE_GROUP_DISCARDED,
+            teacher_log_event_type=TeacherLogEventType.TENTATIVE_GROUP_DISCARDED,
+            comment=self.reason,
```
Auch-Auch, line 27: Note that currently we can not add `group_from` due to `CASCAD` policy at `LogEvent`'s Models

api/views/group.py
```diff
@@ -67,3 +73,34 @@ def confirm_ready_to_start(self, request: Request, pk: int) -> Response:  # noqa
...
+            status.HTTP_400_BAD_REQUEST: OpenApiResponse(
+                response=ValidationErrorSerializer,
+                description="Something is wrong with the data",
+            ),
+        },
+    )
+    @action(detail=True, methods=["delete"])
+    def discard(self, request: Request, pk: int) -> Response:  # noqa: ARG002
+        group = self.get_object()
+        if group.project_status != GroupProjectStatus.PENDING:
```
Auch-Auch, line 101: I also assumed that we can discard it only if a group in the pending state

tests/fixtures/group.py
```diff
@@ -51,3 +51,10 @@ def active_group(timestamp, availability_slots):
     group.coordinators_former.clear()
     group.save()
     yield group
+
+
+@pytest.fixture
+def pending_group(active_group):
+    active_group.project_status = GroupProjectStatus.PENDING
```
Auch-Auch, line 58: I also not sure about this, if do not use a active group my tests fail because `students/teachers/coordinators` filtered by active statuses so I assumed that if someone in the group he/she is in an active status though this might be not true for pending groups

api/serializers/group/internal.py
```diff
@@ -39,3 +42,7 @@ class GroupReadSerializer(serializers.ModelSerializer[Group]):
     class Meta:
         model = Group
         fields = "__all__"
+
+
+class GroupDiscardSerializer(serializers.Serializer[Any]):
```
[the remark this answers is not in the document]
Roxe322: Seems good

api/serializers/group/internal.py
```diff
@@ -39,3 +42,7 @@ class GroupReadSerializer(serializers.ModelSerializer[Group]):
...
+    discard_reason = serializers.ChoiceField(choices=[(e.value) for e in GroupDiscardReason])
```
Roxe322, line 48: I think round brackets are not needed around e.value

api/views/mixins/group.py
```diff
@@ -16,3 +24,36 @@ def create(self, request: Request, *args: tuple[Any], **kwargs: dict[str, Any])
         group = Group.objects.get(id=response.data["id"])
         GroupProcessor.create(group)
         return response
+
+
+class DiscardGroupMixin:
+    @extend_schema(
+        parameters=[
+            OpenApiParameter(
```
Roxe322, line 32: Why do you pass it as a query parameter, not as a body? Then also you could use this seralizer in schema, without manually defining params here

Roxe322: Even with params you could do it, as I see by the docs. IDK why we didn't use it before :) 
(same thing in check_existence_of_chat_id action in personal_info viewset)

https://drf-spectacular.readthedocs.io/en/latest/customization.html#step-2-extend-schema

api/processors/actions/group/discard.py
```diff
@@ -0,0 +1,62 @@
...
+        self.group.delete()
+
+    def _set_group_status(self) -> None:
+        super()._set_group_status()
+
+    def _set_coordinators_status(self) -> None:
+        StatusSetter.update_statuses_of_active_coordinators(self.timestamp)
+
+    def _set_teachers_status(self) -> None:
+        teachers = Teacher.objects.filter_active()
```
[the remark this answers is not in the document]
Roxe322: I'm also not sure which statuses should be set to which teachers, need power of @lemontree210 here

api/processors/actions/group/discard.py
```diff
@@ -0,0 +1,62 @@
...
+    def _create_log_events(self) -> None:
+        GroupLogEventCreator.create(
+            group=self.group,
+            student_log_event_type=StudentLogEventType.TENTATIVE_GROUP_DISCARDED,
+            teacher_log_event_type=TeacherLogEventType.TENTATIVE_GROUP_DISCARDED,
+            comment=self.reason,
+        )
+
+    def _delete(self) -> None:
+        self.group.delete()
```
Roxe322, line 31: Do we really want to completely delete group? Not to set some status as "deleted", or some flag, etc. 

It's better for history of changes.

@lemontree210

api/views/mixins/group.py
```diff
@@ -16,3 +24,36 @@ def create(self, request: Request, *args: tuple[Any], **kwargs: dict[str, Any])
...
+            status.HTTP_400_BAD_REQUEST: OpenApiResponse(
+                response=ValidationErrorSerializer,
+                description="Something is wrong with the data",
+            ),
+        },
+    )
+    @action(detail=True, methods=["delete"])
+    def discard(self, request: Request, pk: int) -> Response:  # noqa: ARG002
+        group = self.get_object()  # type: ignore
+        if group.project_status != GroupProjectStatus.PENDING:
```
Roxe322, line 54: Maybe let's put this logic to serializer's `validate` method?

tests/tests_api/test_group.py
```diff
@@ -477,3 +477,75 @@ def test_dashboard_group_confirm_ready_to_start_general_check(
...
+            list(pending_group.coordinators.iterator()),
+        )
+        students_group_count = [student.groups.count() - 1 for student in group_students]
+        teachers_group_count = [teacher.groups.count() - 1 for teacher in group_teachers]
+        coordinators_group_count = [
+            coordinator.groups.count() - 1 for coordinator in group_coordinators
+        ]
+        discard_reason = GroupDiscardReason.NOT_ENOUGH_STUDENTS
+        response = api_client.delete(
+            f"{base_path}/{pending_group.id}/discard/?discard_reason={discard_reason}"
```
Roxe322, line 505: It's better to pass params with special argument, not as a string, kind of

`api_client.delete(path, {"discard_reason": discard_reason"})`

api/processors/actions/group/discard.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def _set_group_status(self) -> None:
+        super()._set_group_status()
```
Roxe322, line 34: Do we need to redefine this method if it only calls `super()`?

api/processors/actions/group/discard.py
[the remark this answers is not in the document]
Roxe322: After reading issue, I thinks here status changes should only be applied to group's teachers, but I'm not sure. Also not sure about `abort` :D

api/processors/actions/group/discard.py
Auch-Auch: In general yes, because it is an abstract method and you have to define it but we might just leave pass but I prefer to call super because in some cases it might contain default implementation

tests/tests_api/test_group.py
Auch-Auch: Seems like it does not work for `delete`, I tried data and params as well. Maybe it is a reason to change method, not sure

Auch-Auch:
I resolved conflicts but note that I just replaced erd.png with the newest one which probably deleted changes that was made by @lemontree210, not sure what to do here

api/processors/actions/group/discard.py
Roxe322: Ah, okay

Roxe322:
@Auch-Auch Just run "make recreate-first-migration" one more time, it shouldn't change migration itself, but will recreate erd.png

tests/tests_api/test_group.py
Roxe322: Strange. But okay.
Also, I agree with DELETE method if it really deleting instance. If we will decide not to delete it completely - let's change it to POST

api/processors/actions/group/discard.py
lemontree210: this should change in #209 (will be `SET_NULL` instead of `CASCADE`). I think that your PR might be merged earlier than #209, so maybe add a #TODO to add `group_from` once #209 is merged?

api/processors/actions/group/discard.py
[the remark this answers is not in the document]
lemontree210: > I'm also not sure which statuses should be set to which teachers, need power of @lemontree210 here

what power? are you kidding me 😆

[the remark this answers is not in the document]
lemontree210: > After reading issue, I think here status changes should only be applied to group's teachers, but I'm not sure. Also not sure about `abort` :D

Exactly, here we don't need to change coordinators' statuses because this is just a tentative group and it has no coordinator yet.

`abort` is different because this is already a real group that is listed in coordinator's groups. `abort` removes this group from the set, which warrants recalculation of this coordinator's status.

api/views/group.py
lemontree210: That is correct understanding :)

tests/fixtures/group.py
lemontree210: A pending group is an attempt at creating a "really working" group. So it makes sense that even a pending group is composed of people that are active in the project as a whole. It doesn't make any sense to create a group (and any group is pending in the beginning) with people that are not active within the whole project.

api/processors/actions/group/discard.py
lemontree210: > Do we really want to completely delete group? Not to set some status as "deleted", or some flag, etc.
> 
> It's better for history of changes.
> 
> @lemontree210

I know it sounds reasonable, but with the algorithm we might be creating dozens of tentative groups and I'm not sure if all of them are needed. They will be so many that the admins may not be able to use that archive anyway. We discuss something similar in #209 I think

tests/tests_api/test_group.py
lemontree210: I think that tentative groups can be deleted... Maybe that requires further thinking. Or maybe we should just vote :)

api/processors/actions/group/discard.py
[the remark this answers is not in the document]
Auch-Auch: I think the question was about whether to apply statuses to only group's teachers or to the whole set of teachers and the same for abort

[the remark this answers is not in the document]
lemontree210: Oh, OK! Well actually one should only apply statuses to this group's teachers, but we had a discussion with @Roxe322 (I can't remember the exact PR where it was, sorry) and decided that it would be easier to apply the status rules to all teachers. Now, I might be dangerously wrong here, but I don't think it would not lead to unpredictable results (because we're using synchronous Django). And, anyway, a teacher can teach multiple groups and if both groups get aborted, we're getting approximately the same situation.

That being said, yes, the 100% approach would be to only limit the status setters to this group's teachers (and, in case of `abort`, this group's coordinator(s)).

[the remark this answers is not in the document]
Auch-Auch: In that case, here I should filter that if teacher has more groups then his status is working and he has only this group then set status to no group yet, right?

lemontree210:
Correct!

api/processors/actions/group/discard.py
Roxe322: Ok, let's keep it as is and delete them

Auch-Auch (approved):

api/processors/actions/group/discard.py
```diff
@@ -0,0 +1,61 @@
...
+        super().__init__(group)
+
+    @transaction.atomic
+    def process(self) -> None:
+        self._set_statuses()
+        self._create_log_events()
+        self._delete()
+
+    def _create_log_events(self) -> None:
+        # TODO add from_groupd once CASCAD policy removed
```
lemontree210, line 23: @Auch-Auch `CASCADE` was replaced during merging of #209, so you can create a small separate PR to create log events.

PR #222, opened 2023-08-14
Add email field to personal info filter to identify coordinator from dashboard

dont-text-me:
~~Tiny PR to enable filtering of the personal info list by email - to be used in the Tooljet dashboard~~
This PR adds a dashboard viewset to retrieve personal info details - to be used to "authenticate" a coordinator upon first opening the dashboard app
image

### DIFF
--- a/api/serializers/personal_info.py
+++ b/api/serializers/personal_info.py
@@ -54,7 +54,11 @@ class CheckChatIdExistenceSerializer(serializers.ModelSerializer[PersonalInfo]):
 
 
 class DashboardPersonalInfoSerializer(serializers.ModelSerializer[PersonalInfo]):
-    """A serializer used to return information about a person in dashboard (Tooljet) views."""
+    """
+    A serializer used to return information about a person in dashboard (Tooljet) views.
+    The data returned is nested inside a parent entity (teacher or student),
+    so fields like id and name are not returned here.
+    """
 
     class Meta:
         model = PersonalInfo
@@ -65,3 +69,18 @@ class DashboardPersonalInfoSerializer(serializers.ModelSerializer[PersonalInfo])
             "registration_telegram_bot_chat_id",
             "telegram_username",
         )
+
+
+class DashboardStandalonePersonalInfoSerializer(serializers.ModelSerializer[PersonalInfo]):
+    """
+    Used in the Tooljet dashboard to identify a coordinator by their email.
+    Returns minimal data.
+    """
+
+    class Meta:
+        model = PersonalInfo
+        fields = (
+            "id",
+            "email",
+        )
+        read_only_fields = fields
--- a/api/serializers/personal_info.py
+++ b/api/serializers/personal_info.py
@@ -71,7 +71,7 @@ class DashboardPersonalInfoSerializer(serializers.ModelSerializer[PersonalInfo])
         )
 
 
-class DashboardStandalonePersonalInfoSerializer(serializers.ModelSerializer[PersonalInfo]):
+class DashboardMinifiedPersonalInfoSerializer(serializers.ModelSerializer[PersonalInfo]):
     """
     Used in the Tooljet dashboard to identify a coordinator by their email.
     Returns minimal data.
--- a/api/views/personal_info.py
+++ b/api/views/personal_info.py
@@ -15,7 +15,7 @@ from api.serializers import (
     PersonalInfoSerializer,
 )
 from api.serializers.errors import BaseAPIExceptionSerializer, ValidationErrorSerializer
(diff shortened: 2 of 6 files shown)


tests/tests_api/test_personal_info.py
```diff
@@ -29,20 +29,39 @@ def test_personal_info_create(api_client, fake_personal_info_data):
...
     )
     assert response.status_code == status.HTTP_200_OK
+    assert len(response.json()) == 1
+    # checking some basic attrs (cannot compare two objects directly
+    # because some fields were added during creation)
+    for param in ["first_name", "last_name", "email", "information_source"]:
+        assert response.json()[0][param] == fake_personal_info_data[param]
+
 
+def test_personal_info_get_with_params_returns_200_and_data_with_existing_email(
```
Auch-Auch, line 51: The name is a little bit too big for my taste)

Auch-Auch (approved):

lemontree210:
Thanks for the PR! Please extend the decorator so that `email` is shown in params in Swagger, like `registration_bot_chat_id` is.

api/views/personal_info.py
```diff
@@ -56,6 +56,7 @@ def get_serializer_class(self) -> type[BaseSerializer[PersonalInfo]]:
     @extend_schema(
         parameters=[
             OpenApiParameter(name="registration_telegram_bot_chat_id", type=int, required=True),
+            OpenApiParameter(name="email", type=str, required=False),
```
lemontree210, line 59:
```suggestion
```

Please revert this change, this is the wrong place for this parameter. It is a view for the bot that checks the existence of chat ID :)

lemontree210 (changes requested):

Auch-Auch:
> In general, we decided that for dashboard we use routes with `/api/dashboard/` and dashboard serializers. So we cannot just add `email` field to the viewset that is serving the bot. Also, we should use a dashboard serializer, not bot serializer. I should have more time and brain power to do it a bit later, maybe @Auch-Auch you can help out as you remember our logic of dividing viewsets and serializers.

Do you mean that we need a dashboard viewset here? ah, yep, I overlooked that, we should do the same as `students/teachers/coordinators`, so split it up into internal and dashboard. I try will to do it today or tomorrow evening if nobody does it earlier.

dont-text-me:
ah, thanks for pointing that out! just managed to split things up, now looking like this: 
image

Not sure if there is any point in splitting up the filters too, the email field will just sit unused by the bot

lemontree210 (commented):
Hm, I wonder if creation of a separate serializer class (`DashboardStandalonePersonalInfoSerializer`) is really warranted. I need @Roxe322 's expert opinion on that :)

Roxe322 (approved):

api/serializers/personal_info.py
```diff
@@ -65,3 +69,18 @@ class Meta:
             "registration_telegram_bot_chat_id",
             "telegram_username",
         )
+
+
+class DashboardStandalonePersonalInfoSerializer(serializers.ModelSerializer[PersonalInfo]):
```
Roxe322, line 74: Why "Standalone"? Maybe let's rename it to something like
`MinifiedDashboardPersonalInfoSerializer` as we have in other serializers?

tests/tests_api/test_personal_info.py
```diff
@@ -29,26 +29,55 @@ def test_personal_info_create(api_client, fake_personal_info_data):
...
+    assert result["id"] == expected_id
+    assert result["email"] == expected_email
+
+
+def test_dashboard_personal_info_get_handles_unknown_email(api_client, fake_personal_info_list):
+    for item in fake_personal_info_list:
+        api_client.post("/api/personal_info/", data=item)
+
+    response = api_client.get(
+        path="/api/dashboard/personal_info/?email=unknownEmail@samanthasgroup.com",
```
Roxe322, line 75:
```
api_client.get(
    "/api/dashboard/personal_info/",
    {"email": "unknownEmail@samanthasgroup.com"}
)
```

Roxe322: same above

Roxe322 (changes requested):

Roxe322:
Totally agree with adding "minified" serializer with only needed fields @lemontree210

Roxe322 (approved):

dont-text-me:
@lemontree210 i'll let you do the honours once youre happy with the pr👌

lemontree210:
> @lemontree210 i'll let you do the honours once youre happy with the pr👌

Wait, what? 😆 

image

lemontree210 (approved):
the honours are mine :)

PR #228, opened 2023-08-18
#192: Fix model validation

### LINKED ISSUE #192: BUG: admin site gives error when editing a `PersonalInfo` entry
The entry is changed OK, but the error message is shown as if we were trying to create a new record.

Screenshot from 2023-07-24 14-13-01

Roxe322:
Closes #192


lemontree210 (approved):
Thank you for the fix!

PR #229, opened 2023-08-18
#161: Show enrollment test results in students admin

### LINKED ISSUE #161: Admin site: Show student's answers to enrollment test (`EnrollmentTestQuestionOption`s chosen by `Student`, linked to `EnrollmentTestQuestion`)
Admins should be able to view (from student's profile) the answers to the enrollment test that student gave during registration in the bot. These answers can be linked to the questions and/or questions (`EnrollmentTestQuestion.text`) themselves can be listed next to the answers.

Note that at this point, only students that want to study English and can read in English are given the enrollment test. This means that it's OK that some students won't have any `EnrollmentTestQuestionOption` records associated with them.

Roxe322:
Closes #161


api/admin.py
```diff
@@ -39,16 +40,65 @@ class PersonalInfoAdmin(admin.ModelAdmin[models.PersonalInfo]):
...
+
+
+class StudentAdmin(admin.ModelAdmin[models.Student]):
+    form = StudentAdminForm
+    readonly_fields = (
+        "enrollment_tests_summary",
+        "enrollment_tests_result_answers",
+    )
+
+    # TODO: searching is case-sensitive, better to make it case-insensitive somehow
```
lemontree210, line 56: As far as I remember, we automatically normalize names and emails. That would mean the we don't have to make the search case-insensitive.

lemontree210 (approved):

api/admin.py
Roxe322: Yes we do, but it doesn't make search easier. 
For example: we have student Ivan Petrov - you can't find him with text "ivan", only "Ivan", as it's stored in database.

PR #235, opened 2023-08-27
View for finishing studies of the group

### LINKED ISSUE #234: API view: finish studies in a group
When a coordinator from dashboard (bot can't do this) sends a request stating that the studies in the group have finished.

> **Important** (for a separate issue):
>
> Note that afterwards, the coordinator (or the bot? but probably not in MVP) has to communicate with the students and teacher(s) and based on this communication trigger other events (e.g. "awaiting offer" or "finished and leaving")

## Change objects

- move coordinator(s), students and teacher(s) to `..._former` fields of this group.

Since this is a real group, not a tentative one that never studied, we **do not delete** the group unlike in #221.

## Create log events

- coordinator(s): `GROUP_FINISHED` (**add to `CoordinatorLogEventType`** right after `GROUP_ABORTED`)
- group: `FINISHED`
- students: `GROUP_FINISHED`
- teacher(s): `GROUP_FINISHED`

## Remove log event types

**Remove this `StudentLogEventType`**:

```py
    FINISHED_AND_STAYING = (
        "finished_stays",
        "Completed the course and wants to join another group",
    )
```
The students are not supposed to stay in the project once they're done studying, only volunteers are.

## Set `project_status`
- group: `FINISHED`
- Its coordinators' statuses change depending on the number of groups each of them has:
  - if they now have fewer groups than required: `WORKING_BELOW_THRESHOLD`
  - if they still have more groups than the upper limit: `WORKING_LIMIT_REACHED`
  - else: `WORKING_OK`
- students: `STUDYING`
- teachers' statuses depend on their current number of groups:
  - if they have `0` groups:  `NO_GROUP_YET`
  - if they still have groups after this one has finished: `WORKING`

Auch-Auch:
Closes #234


api/models/group.py
```diff
@@ -122,11 +122,15 @@ def __str__(self) -> str:
             f"teachers: {teacher_names}, {self.students.count()} students."
         )
 
-    def teachers_with_other_groups(self) -> QuerySet[Teacher]:
-        return self.teachers.annotate(group_count=Count("groups")).filter(group_count__gt=1)
+    def _active_group_count_annotation(self) -> QuerySet[Teacher]:
+        active_filter = Q(groups__project_status__in=GroupProjectStatus.active_statuses())
```
Auch-Auch, line 126: Please, pay attention to this, I realized that we should count only `active` groups because teacher might have tons of already finished groups.

api/processors/actions/group/finish.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self.group.teachers_with_other_active_groups().update(
+            project_status=TeacherProjectStatus.NO_GROUP_YET,
+            situational_status="",
+            status_since=self.timestamp,
+        )
+
+    def _set_students_status(self) -> None:
+        self.group.students.update(
+            project_status=StudentProjectStatus.STUDYING,
```
Auch-Auch, line 73: Btw, @lemontree210  are you sure it should be `studying`?

lemontree210: Thanks for the attentiveness! I've been thinking about this. Theoretically it is possible that the student is studying in another group. OK this can be checked, but that is not all. If not, we don't know (at this point, when the group is just marked as finished) whether the student is going to leave the project or continue in a more advanced group. This will have to be clarified by the coordinator. So we can either set `STUDYING` or maybe we could add a `project_status` called something like `AWAITING_NEXT_STEP`, which would clearly mark the student as being in a sort of "limbo" in terms of whether they're leaving or should go back to `AWAITING_OFFER` status again, maybe after the coordinator adjusts their language level.

api/models/group.py
lemontree210: Hmm no actually I think they won't: the former groups will be in a teacher's `groups_former` attribute (see `related_name`):

```py
    teachers_former = models.ManyToManyField(
        Teacher,
        blank=True,
        related_name="%(class)ss_former",
        verbose_name="Former teachers",
        help_text=(
            "Lists teachers that once worked with this group, and/or the teacher(s) of this group "
            "when it finished classes or was aborted."
        ),
    )
```

Auch-Auch: Hmm, yes, then it does not change anything, I would still keep it as that to insure that invariant but on the contrary it might confuse people in some point, not sure

api/processors/actions/group/finish.py
Auch-Auch: Get it, then we should keep it studying and mark `finished` once the student leaves permanently

api/models/group.py
lemontree210: I would suggest removing it to avoid confusion. No former group must remain in teacher's `.groups()`, and applying a filter might mask a coding mistake.

api/processors/actions/group/finish.py
lemontree210: On the other hand, you can filter by `AWAITING_NEXT_STEP` and/or put these students somewhere prominently on the dashboard. Of course if we set `STUDYING`, we can display on the dashboard the students that have project_status `STUDYING` but no groups. My point is, however, that no matter which way we choose, we will still need to keep track of these students.

lemontree210: Let's ask the dashboard man @dont-text-me :) What do you think?

api/processors/actions/group/finish.py
```diff
@@ -0,0 +1,76 @@
...
+            self.group.students,
+            self.group.coordinators,
+        )
+        self.group.teachers_former.add(*teachers_current.all())
+        self.group.students_former.add(*students_current.all())
+        self.group.coordinators_former.add(*coordinators_current.all())
+        self.group.teachers.clear()
+        self.group.students.clear()
+        self.group.coordinators.clear()
+        self.group.save()
```
lemontree210, line 48: this is exactly the same in `abort.py`. Maybe this functionality could be factored out to the base class?

api/processors/actions/group/finish.py
```diff
@@ -0,0 +1,76 @@
...
+            project_status=TeacherProjectStatus.WORKING,
+            situational_status="",
+            status_since=self.timestamp,
+        )
+
+        self.group.teachers_with_no_other_groups().update(
+            project_status=TeacherProjectStatus.NO_GROUP_YET,
+            situational_status="",
+            status_since=self.timestamp,
+        )
```
lemontree210, line 69: this is the same in `abort.py` and `discard.py`

tests/tests_api/test_group.py
```diff
@@ -549,3 +549,64 @@ def test_dashboard_group_discard_general_check(
...
+        for teacher in active_group.teachers_former.iterator():
+            assert teacher.project_status in (
+                TeacherProjectStatus.WORKING,
+                TeacherProjectStatus.NO_GROUP_YET,
+            )
+            assert teacher.status_since == common_status_since
+
+            log_event: TeacherLogEvent = TeacherLogEvent.objects.get(teacher_id=teacher.pk)
+            assert log_event.type == TeacherLogEventType.GROUP_FINISHED
+            assert_date_time_with_timestamp(log_event.date_time, timestamp)
```
lemontree210, line 612: Since there is a `test_abort_appends_former_entity_lists()`, maybe we should add a similar test to `finish` as well.

lemontree210 (changes requested):
We can leave refactoring for a separate PR, of course

api/processors/actions/group/finish.py
Auch-Auch: This one indeed can be moved

api/processors/actions/group/finish.py
Auch-Auch: I do not think it is necessary in this case, the main part is already generic and the rest is just status set which is small and more readable as is, I think

lemontree210: > I do not think it is necessary in this case, the main part is already generic and the rest is just status set which is small and more readable as is, I think

Alright :)

api/processors/actions/group/finish.py
dont-text-me: good question! i think we can make do without a limbo status and just go off the combination of the most recent log event and the project status to filter out students that have just "graduated". In terms of presenting it, the simplest and most MVP friendly approach would be to create a new tab in the dashboard (see super high quality mockup attached 😄 ) that would get the list of students already filtered from the backend, since log events are not sent to the dashboard yet. In the future though, i definitely see us wanting a separate page for all the things needing attention displayed as separate widgets, but that's a bit beyond the MVP
image

dont-text-me: so overall, the simplest action we could take from this is to add a query parameter to the `/dashboard/students` view (smth like `groupRecentlyFinished`) that if set, will return filter out students that have `GROUP_FINISHED` as their latest log event + a `STUDYING` project status. I'll then create a tab in the students table to display them all

Auch-Auch: Should we do it in a separate PR or here @lemontree210?

lemontree210: > Should we do it in a separate PR or here @lemontree210?

No limbo status then and I'd say a separate PR (and I'll think about the task a bit more). Thanks @dont-text-me :)

tests/tests_api/test_group.py
```diff
@@ -549,3 +549,84 @@ def test_dashboard_group_discard_general_check(
...
             assert log_event.comment == discard_reason
             assert_date_time_with_timestamp(log_event.date_time, timestamp)
+
+
+class TestDashboardGroupFinish:
+    @staticmethod
+    def _make_url(group: Group) -> str:
+        return reverse("groups-finish", kwargs={"pk": group.id})
+
+    def test_dashboard_group_abort_general_check(self, api_client, active_group, timestamp):
```
lemontree210, line 559:
```suggestion
    def test_dashboard_group_finish_general_check(self, api_client, active_group, timestamp):
```

tests/tests_api/test_group.py
```diff
@@ -549,3 +549,84 @@ def test_dashboard_group_discard_general_check(
...
+                TeacherProjectStatus.WORKING,
+                TeacherProjectStatus.NO_GROUP_YET,
+            )
+            assert teacher.status_since == common_status_since
+
+            log_event: TeacherLogEvent = TeacherLogEvent.objects.get(teacher_id=teacher.pk)
+            assert log_event.type == TeacherLogEventType.GROUP_FINISHED
+            assert_date_time_with_timestamp(log_event.date_time, timestamp)
+
+    def test_abort_appends_former_entity_lists(self, api_client, active_group, availability_slots):
```
lemontree210, line 614:
```suggestion
    def test_finish_appends_former_entity_lists(self, api_client, active_group, availability_slots):
```

lemontree210 (changes requested):
just the naming left :)

lemontree210 (approved):

api/processors/actions/group/finish.py
lemontree210: Created issue #237

PR #240, opened 2023-09-02
feat(AAGC): implement group building algorithm structure

### LINKED ISSUE #252: AAGC: Create structure of group building algorithm

paulin-mipt:
closes #252 

This PR leaves out prioritisation of the group candidates - this piece of logic is crucial and maybe it's better to be reviewed and tested separately.

Also it doesn't attempt to add age ranges on the fly.


api/management/commands/create_groups.py
```diff
@@ -17,15 +22,21 @@ def add_arguments(self, parser: CommandParser) -> None:
         parser.add_argument("--teacher_ids", nargs="*", type=int)
 
     def handle(self, *_: str, **options: Any) -> None:
-        teachers: Iterable[Teacher] = []
+        teachers: Iterable[Teacher]
```
lemontree210, line 25: hm... I don't think this line has any effect now. Does it?

api/management/commands/create_groups.py
```diff
@@ -17,15 +22,21 @@ def add_arguments(self, parser: CommandParser) -> None:
...
         if options["teacher_ids"]:
-            teachers = Teacher.objects.filter(pk__in=options["teacher_ids"])
-            for teacher in teachers:
+            teachers = []
```
lemontree210, line 27: the variable has the same name as the one above. maybe `teachers_for_group_candidate`?

api/management/commands/create_groups.py
```diff
@@ -17,15 +22,21 @@ def add_arguments(self, parser: CommandParser) -> None:
         parser.add_argument("--teacher_ids", nargs="*", type=int)
 
     def handle(self, *_: str, **options: Any) -> None:
-        teachers: Iterable[Teacher] = []
+        teachers: Iterable[Teacher]
         if options["teacher_ids"]:
-            teachers = Teacher.objects.filter(pk__in=options["teacher_ids"])
-            for teacher in teachers:
```
lemontree210, line 23: maybe factor out this loop into a small function to remove the loop from within the if? Just for the sake of readability and maybe also to have operations of the same level of granularity in a function (the else-clause contains a call to a method)

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
-    def _get_group_candidate(teacher_id: int) -> GroupCandidate | None:
-        # THIS IS A STUB METHOD RETURNING DUMMY VALUES
-        # TODO filter for language+level, communication language, age groups, time slots (all UTC?)
-        # TODO iterate in correct priority order, yield results
+    def _get_best_group_candidate(teacher_id: int) -> GroupCandidate | None:
         teacher = Teacher.objects.get(pk=teacher_id)
+        group_candidates: list[GroupCandidate] = []
+
+        for language_and_level in teacher.teaching_languages_and_levels.all():
+            logger.info(language_and_level)
```
lemontree210, line 158:
```suggestion
            logger.debug(language_and_level)
```
just to be consistent with the levels you put in other places

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+                        day_time_slots=(first_time_slot, second_time_slot),
+                    )
+                    if group_candidate is not None:
+                        group_candidates.append(group_candidate)
+
+        if len(group_candidates) == 0:
+            return None
+
+        group_candidates.sort(key=lambda candidate: candidate.priority, reverse=True)
+        return group_candidates[0]
```
lemontree210, line 174:
```suggestion
        return sorted(group_candidates, key=lambda candidate: candidate.priority).pop() if group_candidates else None
```

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+
+    @staticmethod
+    def _get_group_candidate(
+        teacher: Teacher,
+        age_range: AgeRange,
+        language_and_level: LanguageAndLevel,
+        day_time_slots: Collection[DayAndTimeSlot],
```
lemontree210, line 181: It's `Collection` here and `Iterable` elsewhere

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+    ) -> GroupCandidate | None:
+        restrictions = GroupBuilder._get_allowed_group_size(age_range)
+        communication_language = CommunicationLanguageMode(
+            teacher.personal_info.communication_language_mode
+        )
```
lemontree210, line 186: hmm... can you really do that? Or maybe you meant
```py
 communication_language: CommunicationLanguageMode = teacher.personal_info.communication_language_mode
```

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+        students = GroupBuilder._get_students(
+            language_and_level=language_and_level,
+            age_range=age_range,
+            communication_language=communication_language,
+            day_time_slots=day_time_slots,
+            max_students_num=restrictions.max,
+        )
+        logger.debug(students)
+        if len(students) < restrictions.min:
+            return None
```
lemontree210, line 196:
```suggestion
        if len(students) < restrictions.min:
            logger.debug("Not enough students")
            return None
```
I know it's strange to insert debug commands, but I couldn't help it :)

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
         )
+        for time_slot in day_time_slots:
+            student_manager = student_manager.filter(availability_slots=time_slot)
+
+        return student_manager.order_by("personal_info__date_and_time_added")[:max_students_num]
+
+    @staticmethod
+    def _iterate_lesson_times(teacher: Teacher) -> Iterator[tuple[DayAndTimeSlot, DayAndTimeSlot]]:
+        """
+        Iterates tuples of available time slots.
```
lemontree210, line 231:
```suggestion
        Iterate over tuples of available time slots.
```

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+                    first_availability, second_availability
+                ):
+                    yield (first_availability, second_availability)
+
+    @staticmethod
+    def _availability_slots_have_break(
+        first_availability: DayAndTimeSlot, second_availability: DayAndTimeSlot
+    ) -> bool:
+        """
+        Checks that second availability slot is not too early from first
```
lemontree210, line 247:
```suggestion
        Check that second availability slot is not too soon after first
```

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+        """
+        first_day = first_availability.day_of_week_index
+        second_day = second_availability.day_of_week_index
+        if second_day - first_day < MIN_DAYS_BETWEEN_LESSONS:
```
lemontree210, line 251: I think either the constant must be `2` or it should be `<=` instead of `<` here
Also maybe we shouldn't count on the days being ordered and either sort them or compare `abs()`

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+            return False
+        # this checks e.g. Monday and Sunday is just 1 day apart
+        if (first_day + 7) - second_day < MIN_DAYS_BETWEEN_LESSONS:
+            return False
+        return True
+
+    @staticmethod
+    def _get_datetime_kwargs(day_time_slots: Iterable[DayAndTimeSlot]) -> dict[str, time]:
+        datetime_kwargs = {}
```
lemontree210, line 260: Please add a docstring :)

lemontree210 (commented):
Thank you! I'll try and do a second review later today, with emphasis on the logic.

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+                        group_candidates.append(group_candidate)
+
+        if len(group_candidates) == 0:
+            return None
+
+        group_candidates.sort(key=lambda candidate: candidate.priority, reverse=True)
+        return group_candidates[0]
+
+    @staticmethod
+    def _get_group_candidate(
```
lemontree210, line 177:
```suggestion
    def _create_group_candidate(
```
I would rename the method to make it clear that we're not getting this entity from anywhere: we're creating it

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+
+    @staticmethod
+    def _iterate_lesson_times(teacher: Teacher) -> Iterator[tuple[DayAndTimeSlot, DayAndTimeSlot]]:
+        """
+        Iterates tuples of available time slots.
+        There should be at least one free day between lessons.
+        """
+        teacher_availability_slots = teacher.availability_slots.all()
+        for first_availability in teacher_availability_slots:
+            for second_availability in teacher_availability_slots:
```
Roxe322, line 236: Good place to use itertools.product:

```python
for first_availability, second_availabilty in itertools.product(teacher_availability_slots):
    ....
```

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+        Checks that second availability slot is not too early from first
+        """
+        first_day = first_availability.day_of_week_index
+        second_day = second_availability.day_of_week_index
+        if second_day - first_day < MIN_DAYS_BETWEEN_LESSONS:
+            return False
+        # this checks e.g. Monday and Sunday is just 1 day apart
+        if (first_day + 7) - second_day < MIN_DAYS_BETWEEN_LESSONS:
+            return False
+        return True
```
Roxe322, line 256: Doesn't all of this is equal to 
`return abs(first_day - second_day) >= MIN_DAYS_BETWEEN_LESSONS`

api/processors/services/group_builder.py
```diff
@@ -146,23 +150,120 @@ def _get_allowed_group_size(age_range: AgeRange) -> GroupSizeRestriction:
...
+    ) -> bool:
+        """
+        Checks that second availability slot is not too early from first
+        """
+        first_day = first_availability.day_of_week_index
+        second_day = second_availability.day_of_week_index
+        if second_day - first_day < MIN_DAYS_BETWEEN_LESSONS:
+            return False
+        # this checks e.g. Monday and Sunday is just 1 day apart
+        if (first_day + 7) - second_day < MIN_DAYS_BETWEEN_LESSONS:
```
Roxe322, line 254: Maybe something like
```python
days_between_two_weekdays = min((second_day - first_day) % 7, (first_day - second_day) % 7)

return days_between_two_weekdays >= MIN_DAYS_BETWEEN_LESSONS
```

api/processors/services/group_builder.py
lemontree210: (nit)

api/processors/services/group_builder.py
lemontree210: (nit)

api/processors/services/group_builder.py
lemontree210: (nit)

api/processors/services/group_builder.py
lemontree210: (nit)

api/processors/services/group_builder.py
lemontree210: (nit)

api/processors/services/group_builder.py
lemontree210: (nit)

api/processors/services/group_builder.py
lemontree210: (nit: `group_candidates` is created within the function, so it's not crucial to use `sorted()` instead of `.sort`)

api/processors/services/group_builder.py
lemontree210: (nit)

api/management/commands/create_groups.py
lemontree210: (nit, although I do like same level of granularity 😆)

api/management/commands/create_groups.py
lemontree210: this may be irrelevant

api/processors/services/group_builder.py
lemontree210: UPD see [this comment](https://github.com/samanthasgroup/django-webapps/pull/240#discussion_r1320614245) by @Roxe322 instead

api/management/commands/create_groups.py
paulin-mipt: I agree here, changing this. It will also remove that weird type annotation you mentioned on top (yes, it made the difference for one of the linters, apparently)

api/processors/services/group_builder.py
paulin-mipt: Yes, sorry, here my personal opinion is about pop() being less readable in a one'liner. I'll stick to it here. It's a neverending discussion about "less lines means more readable or not?" and in many cases there's no true right and wrong :)

api/processors/services/group_builder.py
paulin-mipt: `get` is really not the right one, agree. But I'd rename to `build` instead. In many cases `create` has semantics of saving something permanently, and here it's just  temporary object construction. If we're being pedantic, that would be my take :D

api/processors/services/group_builder.py
paulin-mipt: In a data model `communication_language_mode` is a string and not `CommunicationLanguageMode`. I prefer using actual enumeration objects instead of values where possible. And Django choices are a slightly sophisticated enum object. That's what you can actually do - get a enum object from its value using this bracket syntax. So you'll be surprised but it is valid (like in [this SO](https://stackoverflow.com/questions/38716288/get-enumeration-name-by-value))

api/processors/services/group_builder.py
paulin-mipt: Errrm, that's just a lot of logs. Imagine we try every possible combination of language, level, date time and most of them don't work (maybe not on real data in fact! But in our testing database it's a bit of pain to find someting useful :cry: )

api/processors/services/group_builder.py
paulin-mipt: I argue if that's more readable but I'll give you that. I know there's rarely a moment when one can legitimately use itertools!

api/processors/services/group_builder.py
paulin-mipt: Good job spotting `<` instead of `<=`! As for `abs` - actually that was point to also filter out second <= first pairs. Added it to docstring to clarify

api/processors/services/group_builder.py
paulin-mipt: Cool, thank you! Had to go to python shell to make sure I remember how % operator works on negatives. Guess I need to retake primary school maths course :D

Although had to change `>=` to `>` because there was a bug initially, which @lemontree210 spotted

api/processors/services/group_builder.py
lemontree210: TIL

lemontree210 (approved):
LGTM

api/processors/services/group_builder.py
lemontree210: > I argue if that's more readable but I'll give you that. I know there's rarely a moment when one can legitimately use itertools!

[I'll give you that](https://www.youtube.com/watch?v=R7GeKLE0x3s&t=665s) 😆

PR #241, opened 2023-09-03
Feature/student transfer

### LINKED ISSUE #239: API view: transfer student from group to group
When a coordinator (from dashboard) stating that student `X` needs to be transferred from group `Y` to group `Z`:

## Create log events

- student: `TRANSFERRED` with `from_group` set to group `Y` and `to_group` set to group `Z`
- no other log events need to be created

## Create/change other objects
- move student from `.students` to `.students_former` in group `Y`
- add this student to `.students` in group `Z`

## Set statuses

No `project_status`es need to be changed. The student can only be transferred from group to group if they are `STUDYING` already. If they are `ON_LEAVE`, they should not be a member of any group. In all other cases of student's possible `project_status`, it's even clearer that the student cannot have them and be transferred from group to group.

`situational_status` must be explicitly set to `""`. The `NEEDS_TRANSFER` situational status is for the case when the student requested a transfer and they should be marked as requiring a transfer for better visibility of such students. But when a transfer is complete (which is our case here), the `situational_status` becomes blank since the situation right after the transfer is contradictory to any of situational statuses we now have for our students.

Auch-Auch:
Closes #239


Auch-Auch:
@lemontree210 Since there is no concrete description of the issue I just added initial functionality based on my assumptions.

lemontree210:
> @lemontree210 Since there is no concrete description of the issue I just added initial functionality based on my assumptions.

OMG I'm losing my concentration. Sorry! Obviously, you know a lot about business logic by now, but it's still sloppy of me. Will update the description.

api/processors/actions/student/transfer.py
```diff
@@ -0,0 +1,50 @@
...
+    def process(self) -> None:
+        self._transfer()
+        self._set_statuses()
+        self._create_log_events()
+
+    def _transfer(self) -> None:
+        self.old_groups: QuerySet[Group] = self.student.groups.all()
+        for old_group in self.old_groups:
+            self.student.groups.remove(old_group)
+            old_group.students_former.add(self.student)
```
dont-text-me, line 26: hmm, this may, of course, be the intended behaviour, but personally, i'd rather ask for a `from_group` or smth in the request body and only remove the student from that group.

what would happen when a student is, for example, studying english A1 in one group and spanish A1 in another and is then transferred (upon their request) to an english B1 group?

api/processors/actions/student/transfer.py
```diff
@@ -0,0 +1,50 @@
...
+    def _set_statuses(self) -> None:
+        self.student.project_status = StudentProjectStatus.STUDYING
+        self.student.situational_status = ""
+        self.student.status_since = self.timestamp
+        self.student.save()
+
+    def _create_log_events(self) -> None:
+        if self.old_groups:
+            for old_group in self.old_groups:
+                StudentLogEvent.objects.create(
```
dont-text-me, line 41: similar to the comment above, i think not asking for the id of the group the student is being transferred from is gonna lead to more of these events than necessary

api/processors/actions/student/transfer.py
```diff
@@ -0,0 +1,50 @@
...
+    @transaction.atomic
+    def process(self) -> None:
+        self._transfer()
+        self._set_statuses()
+        self._create_log_events()
+
+    def _transfer(self) -> None:
+        self.old_groups: QuerySet[Group] = self.student.groups.all()
+        for old_group in self.old_groups:
+            self.student.groups.remove(old_group)
```
lemontree210, line 25: A situation is theoretically possible where a student attends 2 groups simultaneously (e.g. different languages). This operation will remove this student from both groups, not only from the group the student is being transferred from.

same for `._create_log_events()` below.

api/processors/actions/student/transfer.py
```diff
@@ -0,0 +1,50 @@
...
+                StudentLogEvent.objects.create(
+                    student=self.student,
+                    from_group=old_group,
+                    to_group=self.to_group,
+                    type=StudentLogEventType.TRANSFERRED,
+                )
+        else:
+            StudentLogEvent.objects.create(
+                student=self.student, to_group=self.to_group, type=StudentLogEventType.TRANSFERRED
+            )
```
lemontree210, line 50: this is literally impossible: a transfer can only be from one group to another, otherwise it's not a transfer.

lemontree210 (changes requested):
one general question related to business logic

dont-text-me (commented):
looking good so far 👍 , just a small question about the proposed request schema, maybe a question more suited for our leads/product owners as i imagine the case im worried about will happen pretty rarely (if ever)

api/processors/actions/student/transfer.py
dont-text-me: oops, had the same exact comment as @lemontree210 , lets continue in his thread 👌

api/processors/actions/student/transfer.py
Auch-Auch: So it means we need, `transfer_from_id`?

lemontree210: Yes, the API view should require both from_id and to_id to know what group the student is being transferred from and what group they are being transferred to.

Maybe they could be called `from_group_id` and `to_group_id`.

api/serializers/student/dashboard.py
```diff
@@ -65,3 +68,27 @@ class Meta(CommonDashboardStudentSerializer.Meta):
...
+        except Group.DoesNotExist:
+            raise ConflictError("Transfer to group is no found")
+
+        try:
+            from_group = Group.objects.get(pk=int(attrs["from_group_id"]))
+            if self.instance is not None and self.instance not in from_group.students.all():
+                raise ConflictError("Student must be in group from which he is being transferred")
+            attrs["from_group"] = from_group
+        except Group.DoesNotExist:
+            raise ConflictError("Transfer from group is no found")
```
dont-text-me, line 92: definitely more of a nitpick but it would be easier to debug things if these included the ID's of things e.g `Student X must be in group Y` and so on

api/processors/actions/student/transfer.py
```diff
@@ -0,0 +1,41 @@
...
+    def _transfer(self) -> None:
+        self.student.groups.remove(self.from_group)
+        self.from_group.students_former.add(self.student)
+        self.to_group.students.add(self.student)
+        self.student.save()
+        self.to_group.save()
+
+    def _set_statuses(self) -> None:
+        self.student.project_status = StudentProjectStatus.STUDYING
+        self.student.situational_status = ""
```
dont-text-me, line 31: IIRC this is nullable, so you could set it to `None`?

dont-text-me (commented):
LGTM logic-wise, had some nitpicks but overall looks good from my side. Will let the django masters make the final call 👌

api/serializers/student/dashboard.py
lemontree210: Agreed!

api/processors/actions/student/transfer.py
lemontree210: Hm... I don't think a `CharField` can be `null=True`.

```py
    situational_status = models.CharField(
        max_length=DEFAULT_CHOICE_CHAR_FIELD_MAX_LENGTH,
        choices=TeacherSituationalStatus.choices,
        blank=True,
    )
```

dont-text-me: Ah ok, all good then👌

api/views/student.py
```diff
@@ -29,6 +36,34 @@ class DashboardStudentViewSet(viewsets.ReadOnlyModelViewSet[Student]):
     queryset = Student.objects.all()
     serializer_class = DashboardStudentSerializer
 
+    @extend_schema(
+        request=DashboardStudentTransferSerializer,
+        responses={
+            status.HTTP_204_NO_CONTENT: OpenApiResponse(description="Student is transferred"),
```
Roxe322, line 42: You return 200 in line 65 :)

api/serializers/student/dashboard.py
```diff
@@ -65,3 +68,27 @@ class Meta(CommonDashboardStudentSerializer.Meta):
...
         )
+
+
+class DashboardStudentTransferSerializer(serializers.Serializer[Any]):
+    to_group_id = serializers.IntegerField()
+    from_group_id = serializers.IntegerField()
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        try:
+            to_group = Group.objects.get(pk=int(attrs["to_group_id"]))
```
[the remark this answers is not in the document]
Auch-Auch: Reason?

api/views/student.py
Auch-Auch: Good catch

api/serializers/student/dashboard.py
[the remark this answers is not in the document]
Roxe322: Because the meaning of try statement is to cover the spot where exception could occur. If we don't wait an exception in some line(s) of code, why we need to put it into try..except block?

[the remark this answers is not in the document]
Roxe322: It will also make a code more readable, you will remove unnecessary nesting, and it will be easier to understand where exception could be raised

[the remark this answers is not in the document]
Auch-Auch: First point is reasonable, less error prone but in my sense code is more readable when you use try-catch as an abstraction but, yes, in this case you have another exception inside which is somehow confusing

lemontree210 (approved):

PR #244, opened 2023-09-05
Copy `day_and_time_slot` and `language_and_level` views to the dashboard router

### LINKED ISSUE #243: Housekeeping: duplicate the `languages_and_levels` and `day_and_time_slots` views in the dashboard router
These 2 views are needed when creating a new group: the first one enables the list of languages and levels to sync with the database, and the other enables getting the timeslot ID from the provided day of week and time range. 
I have been using the non-dashboard URL's for prototyping, but it seems like we do want to go ahead with this functionality.

dont-text-me:
closes #243

### DIFF
--- a/api/models/auxil/status_setter.py
+++ b/api/models/auxil/status_setter.py
@@ -30,7 +30,7 @@ class StatusSetter:
 
     @staticmethod
     def update_statuses_of_active_coordinators(timestamp: datetime.datetime) -> None:
-        coordinators = Coordinator.objects.filter_active()  # type: ignore[attr-defined]
+        coordinators = Coordinator.objects.filter_active()
 
         coordinators.filter_below_threshold().update(
             project_status=CoordinatorProjectStatus.WORKING_BELOW_THRESHOLD, status_since=timestamp
--- a/api/processors/actions/group/abort.py
+++ b/api/processors/actions/group/abort.py
@@ -43,7 +43,7 @@ class GroupAbortProcessor(GroupActionProcessor):
         )
 
     def _set_teachers_status(self) -> None:
-        teachers = Teacher.objects.filter_active()  # type: ignore[attr-defined]
+        teachers = Teacher.objects.filter_active()
 
         teachers.filter_has_groups().update(
             project_status=TeacherProjectStatus.WORKING,
--- a/api/urls.py
+++ b/api/urls.py
@@ -61,6 +61,12 @@ dashboard_router.register(
     DashboardPersonalInfoViewSet,
     basename="personal_info",
 )
+dashboard_router.register(
+    r"languages_and_levels", LanguageAndLevelViewSet, basename="languages_and_levels"
+)
+dashboard_router.register(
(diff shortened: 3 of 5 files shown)


api/processors/actions/group/abort.py
```diff
@@ -43,7 +43,7 @@ def _set_group_status(self) -> None:
         )
 
     def _set_teachers_status(self) -> None:
-        teachers = Teacher.objects.filter_active()  # type: ignore[attr-defined]
```
dont-text-me, line 46: not sure what happened here, these seem to have been added by a bot and were messing with the mypy check in pre-commit

api/models/auxil/status_setter.py
```diff
@@ -30,7 +30,7 @@ def set_status(
 
     @staticmethod
     def update_statuses_of_active_coordinators(timestamp: datetime.datetime) -> None:
-        coordinators = Coordinator.objects.filter_active()  # type: ignore[attr-defined]
```
dont-text-me, line 33: same here

api/urls.py
```diff
@@ -61,6 +61,12 @@
     DashboardPersonalInfoViewSet,
     basename="personal_info",
 )
+dashboard_router.register(
+    r"languages_and_levels", LanguageAndLevelViewSet, basename="languages_and_levels"
+)
+dashboard_router.register(
+    r"day_and_time_slots", DayAndTimeSlotViewSet, basename="day_and_time_slots"
+)
```
Auch-Auch, line 69: Probably wee need to create a separate view set , e.g: `DashboardDayAndTimeSlotViewSet`, just to be consistent

Auch-Auch (changes requested):

api/urls.py
dont-text-me: hmm, not sure where the difference would actually be, the dashboard needs the exact same format of data and the non-dahsboard views are already read-only which is also what we want

Auch-Auch: Currently there is not difference so it is just to follow the same design approach as we did everywhere. But it might in a different PR of course

api/processors/actions/group/abort.py
lemontree210: Nah, I don't think bots add those, I think @Roxe322 added them because mypy didn't like the filter. Maybe what happened was that mypy got upgraded at started to understand custom filters, hence now complaining about a redundant `type: ignore`.

api/urls.py
lemontree210: Well @dont-text-me since you created this PR to be consistent with the logic of separation of bot-oriented URLs from dashboard-oriented ones, then to me it would be only logical to follow this approach in the viewsets as well. Otherwise it will be a sort of half-PR :)

api/models/auxil/status_setter.py
lemontree210: [yep](https://github.com/samanthasgroup/django-webapps/pull/244#discussion_r1316421442)

api/urls.py
dont-text-me: fair enough 😄

Roxe322 (approved):

PR #253, opened 2023-09-10
Teacher transfer from group to group

### LINKED ISSUE #251: API view: transfer teacher from group to group
same as #239 but for teachers

Auch-Auch:
Closes #251


api/processors/actions/teacher/transfer.py
```diff
@@ -0,0 +1,41 @@
...
+        self.teacher.situational_status = ""
+        self.teacher.status_since = self.timestamp
+        self.teacher.save()
+
+    def _create_log_events(self) -> None:
+        TeacherLogEvent.objects.create(
+            teacher=self.teacher,
+            from_group=self.from_group,
+            to_group=self.to_group,
+            type=TeacherLogEventType.STUDY_START,
```
Auch-Auch, line 40: It is not a final status, for teacher we do not have a `TRANSFERRED` yet.

lemontree210: Yes, please do, maybe along with the others, just to be consistent with student log event types:
```py
    REQUESTED_TRANSFER = "requested_transfer", "Requested transfer"
    TRANSFERRED = "transferred", "Transferred"
    TRANSFER_CANCELED = "transfer_canceled", "Transfer canceled"
```

api/serializers/shared.py
```diff
@@ -0,0 +1,27 @@
...
+
+class PersonTransferSerializer(serializers.Serializer[Any]):
+    to_group_id = serializers.IntegerField()
+    from_group_id = serializers.IntegerField()
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        try:
+            to_group = Group.objects.get(pk=int(attrs["to_group_id"]))
+        except Group.DoesNotExist:
+            raise ConflictError(f"Group {attrs['to_gorup_id']} is no found")
```
lemontree210, line 17:
```suggestion
            raise ConflictError(f"Group {attrs['to_group_id']} not found")
```
I spotted this typo in `StudentTransferSerializer`, but just corrected it before merging. Not sure it's a big deal that no tests broke (should have thrown `KeyError`)

api/serializers/shared.py
```diff
@@ -0,0 +1,27 @@
...
+
+        attrs["to_group"] = to_group
+
+        try:
+            from_group = Group.objects.get(pk=int(attrs["from_group_id"]))
+        except Group.DoesNotExist:
+            raise ConflictError(f"Group {attrs['from_group_id']} is no found")
```
lemontree210, line 24:
```suggestion
            raise ConflictError(f"Group {attrs['from_group_id']} not found")
```

lemontree210 (changes requested):
minor remarks + maybe extend a test?

api/processors/actions/teacher/transfer.py
```diff
@@ -0,0 +1,41 @@
...
+        self._transfer()
+        self._set_statuses()
+        self._create_log_events()
+
+    def _transfer(self) -> None:
+        self.teacher.groups.remove(self.from_group)
+        self.from_group.teachers_former.add(self.teacher)
+        self.to_group.teachers.add(self.teacher)
+        self.teacher.save()
+        self.to_group.save()
```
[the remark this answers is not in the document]
Roxe322: I also think that `teacher.save()` is not needed here

api/serializers/teacher/dashboard.py
```diff
@@ -86,3 +90,20 @@ class Meta(CommonDashboardTeacherSerializer.Meta):
...
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        validated_attrs = super().validate(attrs)
+        to_group = validated_attrs["to_group"]
+        from_group = validated_attrs["from_group"]
+        if self.instance is not None and self.instance in to_group.teachers.all():
+            raise ConflictError(
+                f"Teacher {self.instance.pk} is already in that group {to_group.pk}"
+            )
+
+        if self.instance is not None and self.instance not in from_group.teachers.all():
```
Roxe322, line 105: I think here and above it's better to replace list-finding (`self.instance not in from_group.teachers.all()`) with sql-query (`from_group.teachers.filter(pk=self.instance.pk).exists()`) because of more efficiency

api/serializers/teacher/dashboard.py
```diff
@@ -86,3 +90,20 @@ class Meta(CommonDashboardTeacherSerializer.Meta):
...
+            teacher_id = self.instance.pk
```
Roxe322, line 106: Do we really need this one-time variable? Why not just put self.instance.pk into f-string below?

lemontree210: nit :)

dont-text-me (commented):
LGTM as far as request structure and logic is concerned 👌

api/processors/actions/teacher/transfer.py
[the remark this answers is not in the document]
Auch-Auch: Let's add `.save()` for both group and remove it for teacher since we call it in `_set_statuses`

api/serializers/shared.py
Auch-Auch: `to_group` is always exists in tests so it is good that you catched it at this stage

api/processors/teacher.py
```diff
@@ -0,0 +1,11 @@
...
+from api.models.group import Group
+from api.processors.actions import TeacherTransferProcessor
+
+
+class TeacherProcessor:
+    """A Facade class providing access to actions with teachers."""
+
+    @staticmethod
+    def transfer(teacher: Teacher, to_group: Group, from_group: Group) -> None:
+        TeacherTransferProcessor(teacher, to_group, from_group).process()
```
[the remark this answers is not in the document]
Auch-Auch: `to_group` and `from_group` can not be `None` so it does not really make sense to do it, I think and, yes, the order is like whether you are a right hand or left hand :)

api/serializers/teacher/dashboard.py
Auch-Auch: Yes, no need anymore

api/processors/actions/student/transfer.py
```diff
@@ -23,7 +23,7 @@ def _transfer(self) -> None:
         self.student.groups.remove(self.from_group)
         self.from_group.students_former.add(self.student)
         self.to_group.students.add(self.student)
-        self.student.save()
+        self.from_group.save()
         self.to_group.save()
```
lemontree210, line 27: As far as I understand, `.save()` saves the actual table in the database, while all changes in `._transfer()` affect only m2m relationships, i.e. only secondary tables are changed. If my understanding is correct, then calling `.save()` 2 times makes 2 unnecessary DB queries per each call. (That's why I keep asking this, as this is not about beauty of code, but about DB hits)

I think the easy way is to remove these lines and run the tests to see what happens :)

api/processors/actions/teacher/transfer.py
```diff
@@ -0,0 +1,41 @@
...
+        self._transfer()
+        self._set_statuses()
+        self._create_log_events()
+
+    def _transfer(self) -> None:
+        self.teacher.groups.remove(self.from_group)
+        self.from_group.teachers_former.add(self.teacher)
+        self.to_group.teachers.add(self.teacher)
+        self.to_group.save()
+        self.from_group.save()
```
lemontree210, line 27: same

api/processors/actions/student/transfer.py
Auch-Auch: Well, tests are passed in both cases and I did not even consider such cases, because in my case it is the same as optimizing two loops in one, which in 99% does not make a difference. And here as well, you might have 10-30 members of the group and even if you update them twice it will make no difference, especially with postgres as it has an advanced planner. So here I wanted it to be as much clearer as it can be and include both so that people do not think about internal implementation when reading this code.

Auch-Auch: Btw, not sure I 100% understand why it works only with one `group.save()`, IN one case we add teacher to former and it is not related to other group,  magic :)

lemontree210: > Well, tests are passed in both cases and I did not even consider that, because in my head it is the same as optimizing two loops in one, which in 99% does not make a difference. And here as well, you might have 10-30 members of the group and even if you update them twice it will make no difference, especially with postgres as it has an advanced planner. So here I wanted to be consistent and include both so that people do not think about internal implementation when reading this code.

OK, I see your point. I guess @Roxe322 would argue that people reading the code should either know Django or trust us in that we know Django 😆 

Since I'm starting to go too deep into the discussion, please just let me know if @Roxe322 's comments were addressed, and if so, I'll just approve.

Auch-Auch: Yep, should be done, also added one more test to insure that you can not move to an none existing group (and remove from old one)

lemontree210:
OK. I'm approving then, don't be angry @Roxe322 , I'm just trying to make things go forward (contrary to what I have been doing up until now 😆 )

lemontree210 (approved):

PR #254, opened 2023-09-10
Bug fix: coordinator email filter not being applied in the student with personal info view

dont-text-me:
Just a quick PR, noticed this while testing out the tooljet interface

image


lemontree210 (approved):
Great, thanks! Maybe we could also add a test to make sure the filter for project_status works as well, but it's not 100% necessary I guess.

dont-text-me:
> Great, thanks! Maybe we could also add a test to make sure the filter for project_status works as well, but it's not 100% necessary I guess.

Added in #256

PR #258, opened 2023-09-14
#237: show students who are awaiting coordinator's decision

### LINKED ISSUE #237: API view filter: show students who are awaiting coordinator's decision
The issue originally was that coordinators must be able to easily see students that finished studying so that they can perform one of (at least) two actions: remove them from the project or mark them as awaiting another group (see [discussion](https://github.com/samanthasgroup/django-webapps/pull/235#discussion_r1306652201) in #235 for how @dont-text-me saw implementation for the dashboard). 

However, after [discussion](https://github.com/samanthasgroup/django-webapps/issues/237#issuecomment-1712574001) with @Roxe322, I suggest creating a new `project_status` for `Student`, called `AWAITING_DECISION`, which for now will be applied to students that finished the group (hence a follow-up on #235 will have to be created with amended `project_status`).

Roxe322:
closes #237


api/processors/actions/group/finish.py
```diff
@@ -55,8 +57,22 @@ def _set_teachers_status(self) -> None:
         )
 
     def _set_students_status(self) -> None:
-        self.group.students.update(
+        annotated_students = Student.objects.annotate(groups_count=Count("groups")).filter(
+            groups=self.group
+        )
+
+        # TODO: We should check group status or something
```
Roxe322, line 64: @lemontree210 I need your help here :)

lemontree210: When a group gets finished or aborted, all students are moved to its  `.students_former`, and this group will be in the student's `.groups_former`. This should mean that `.groups` only shows groups that are either working or awaiting start, so I don' think any additional check of group status is needed.

lemontree210: Another thing is that it is strange from the point of readability when we set status `AWAITING_DECISION` for students that have `groups_count=1`. I know that this is because this method is called before the students are moved from `.students` to `.students_former`. But I like the way `._set_teachers_status()` is written in that it is clear that if a teacher has no other groups (not "if teacher has 1 group"), then a certain status should be set.

Of course, you can simply add a comment saying that 
```
# after moving the student, this 1 will become 0
```

Roxe322: Let's just add a comment to be consistent with other solutions :)

lemontree210 (approved):

PR #264, opened 2023-09-17
Student missed class view

### LINKED ISSUE #263: API view: student missed class with vs. without notifying the teacher
When a coordinator (from dashboard) sends a request stating that the given student missed a class:

## Create log events

Types of `LogEvent`s to be created for `Student`:

- `MISSED_CLASS_NOTIFIED` if student notified the teacher in advance
- `MISSED_CLASS_SILENTLY` if student hadn't notified the teacher

Obviously, the backend will know nothing about the fact of notification (or lack thereof), so the coordinator just has to push a respective button (or whatever) in the dashboard (after being notified by the teacher).

Basically this issue is about creating two views in one (e.g. "missed class" view with parameter of "notified" set to true or false).

## Set `situational_status` for `Student` if conditions are met

If `STUDENT_CLASS_MISS_LIMIT` has been reached (e.g. if, together with this one, there is a total of `STUDENT_CLASS_MISS_LIMIT` log events of type `MISSED_CLASS_SILENTLY` for this student over the past 2 weeks), set `situational_status` to `NOT_ATTENDING`.

> [!note]
> Why "over the past 2 weeks"? Because actually the owners want to set status to `NOT_ATTENDING` if 3 classes **in a row** were missed. But there's no way for us of knowing that, because there is no event `ATTENDED_CLASS`. So I suggest saying that if a student missed 3 classes over the past 2 weeks, then with our default frequency of 2 classes per week it should be roughly equivalent to missing 3 classes in a row. We can put `2` (for 2 weeks) in a constant as well, along with a shorter version of this explanation here as its docstring.

> [!important]
> Events of type `MISSED_CLASS_NOTIFIED` don't count towards setting `situational_status` to `NOT_ATTENDING`

Auch-Auch:
Closes #263

### DIFF
--- a/api/processors/actions/student/missed_class.py
+++ b/api/processors/actions/student/missed_class.py
@@ -1,0 +1,47 @@
+from datetime import timedelta
+
+from django.db import transaction
+
+from api.models import Group, Student
+from api.models.auxil.constants import STUDENT_CLASS_MISS_LIMIT
+from api.models.choices.log_event_type import StudentLogEventType
+from api.models.choices.status import StudentSituationalStatus
+from api.models.log_event import StudentLogEvent
+from api.processors.actions.student import StudentActionProcessor
+
+
+class StudentMissedClassProcessor(StudentActionProcessor):
+    def __init__(self, student: Student, group: Group, notified: bool):
+        self.notified = notified
+        self.group = group
+        super().__init__(student)
+
+    @transaction.atomic
+    def process(self) -> None:
+        self._create_log_events()
+        self._set_statuses()
+
+    def _is_student_reached_limit(self) -> bool:
+        last_two_weeks_date = self.timestamp - timedelta(weeks=2)
+        student_class_misses = StudentLogEvent.objects.filter(
+            date_time__gte=last_two_weeks_date, type=StudentLogEventType.MISSED_CLASS_SILENTLY
+        ).all()
+        if student_class_misses.count() >= STUDENT_CLASS_MISS_LIMIT:
+            return True
+        return False
+
+    def _set_statuses(self) -> None:
+        if self._is_student_reached_limit():
+            self.student.situational_status = StudentSituationalStatus.NOT_ATTENDING
+            self.student.status_since = self.timestamp
+            self.student.save()
+
+    def _create_log_events(self) -> None:
+        t = (
+            StudentLogEventType.MISSED_CLASS_NOTIFIED
+            if self.notified
+            else StudentLogEventType.MISSED_CLASS_SILENTLY
+        )
+        StudentLogEvent.objects.create(
+            student=self.student, from_group=self.group, to_group=self.group, type=t
+        )
--- a/api/processors/actions/student/missed_class.py
+++ b/api/processors/actions/student/missed_class.py
@@ -24,7 +24,9 @@ class StudentMissedClassProcessor(StudentActionProcessor):
     def _is_student_reached_limit(self) -> bool:
         last_two_weeks_date = self.timestamp - timedelta(weeks=2)
         student_class_misses = StudentLogEvent.objects.filter(
-            date_time__gte=last_two_weeks_date, type=StudentLogEventType.MISSED_CLASS_SILENTLY
+            date_time__gte=last_two_weeks_date,
+            type=StudentLogEventType.MISSED_CLASS_SILENTLY,
+            student=self.student,
         ).all()
         if student_class_misses.count() >= STUDENT_CLASS_MISS_LIMIT:
             return True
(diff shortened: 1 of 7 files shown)


api/processors/actions/student/missed_class.py
```diff
@@ -0,0 +1,47 @@
...
+    @transaction.atomic
+    def process(self) -> None:
+        self._create_log_events()
+        self._set_statuses()
+
+    def _is_student_reached_limit(self) -> bool:
+        last_two_weeks_date = self.timestamp - timedelta(weeks=2)
+        student_class_misses = StudentLogEvent.objects.filter(
+            date_time__gte=last_two_weeks_date, type=StudentLogEventType.MISSED_CLASS_SILENTLY
+        ).all()
```
lemontree210, line 28: do we have to filter by `student` here also?

api/processors/actions/student/missed_class.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def _create_log_events(self) -> None:
+        t = (
+            StudentLogEventType.MISSED_CLASS_NOTIFIED
+            if self.notified
+            else StudentLogEventType.MISSED_CLASS_SILENTLY
+        )
+        StudentLogEvent.objects.create(
+            student=self.student, from_group=self.group, to_group=self.group, type=t
+        )
```
lemontree210, line 47:
```suggestion
        event_type = (
            StudentLogEventType.MISSED_CLASS_NOTIFIED
            if self.notified
            else StudentLogEventType.MISSED_CLASS_SILENTLY
        )
        StudentLogEvent.objects.create(
            student=self.student, from_group=self.group, to_group=self.group, type=event_type
        )
```
Just a nitpick while we're at it

api/processors/actions/student/missed_class.py
Auch-Auch: Yes

lemontree210 (approved):
LGTM

api/processors/actions/student/missed_class.py
```diff
@@ -0,0 +1,49 @@
...
+        super().__init__(student)
+
+    @transaction.atomic
+    def process(self) -> None:
+        self._create_log_events()
+        self._set_statuses()
+
+    def _is_student_reached_limit(self) -> bool:
+        last_two_weeks_date = self.timestamp - timedelta(weeks=2)
+        student_class_misses = StudentLogEvent.objects.filter(
```
Roxe322, line 26: Do we really need this variable? Why we just don't get count here?

Also below it could be something like 
```
return student_class_misses_count >= STUDENT_CLASS_MISS_LIMIT
```

api/serializers/student/dashboard.py
```diff
@@ -88,3 +89,21 @@ def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
...
+
+class DashboardStudentMissedClassSerializer(serializers.Serializer[Any]):
+    group_id = serializers.IntegerField()
+    notified = serializers.BooleanField()
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        try:
+            group = Group.objects.get(pk=int(attrs["group_id"]))
+        except Group.DoesNotExist:
+            raise ConflictError(f"Group {attrs['group_id']} not found")
```
Roxe322, line 102: Is it really ConflictError? Maybe some other type of exception?

Roxe322:
First comment more about codestyle, I think it's important here, second one is not really

api/serializers/student/dashboard.py
Auch-Auch: We have discussed it at some point I think so if we use some param as a part of the URL (`api/v2/user/<id>`) then it might be a good idea to use not found, on the contrary, if something inside a body it means that we found a source but due to some conflicts were unable to return it. At least I  follow this logic

api/processors/actions/student/missed_class.py
Auch-Auch: Yes, it might make sense to use count instead of all if I got you correctly

api/serializers/student/dashboard.py
Roxe322: But here we didn't really find anything, so it seems that NotFound is more suitable. But if in other places we use ConflictError for such cases - let's keep it as it is

api/processors/actions/student/missed_class.py
Roxe322: Yeah, and also return statement instead of 
```
if x: 
    return True
else:
    return False
``` 
it's easier to do just

```
return x
```

api/serializers/student/dashboard.py
lemontree210: Let me interfere here @Roxe322 , we had this discussion with you and looked at a huge thread on SO. The main argument against 404 is that in our case the address is valid and the server can respond correctly (just has nothing to show), while 404 gives the signal that the URL is completely wrong, no resource was found.

api/processors/actions/student/missed_class.py
lemontree210: Just don't get bogged down in code style issues @Roxe322 , remember that I suggested that I remain the only one (the youngest junior 😆 ) that gets reviewed on this level :)

lemontree210: @Auch-Auch you can just merge whenever you feel you have addressed this comment

PR #273, opened 2023-09-24
Add went on leave view  for teacher and student

### LINKED ISSUE #267: API view: `Student` or `Teacher` went on leave
## Add log event types before creating serializers and views

Please add the following log event types to `StudentLogEventType`, `TeacherLogEventType` and `TeacherUnder18LogEventType` (they can be copied from `CoordinatorLogEventType`)

```py
    GONE_ON_LEAVE = "gone_on_leave", "Gone on leave"
    RETURNED_FROM_LEAVE = "returned_from_leave", "Returned from leave"
```

When a coordinator from dashboard or bot sends a request stating that a student or a teacher went on leave:

## Create log events

Type of `LogEvent`s to be created: `GONE_ON_LEAVE`

## Set statuses
`project_status` is to be set to `ON_LEAVE`.

> [!Important]
> Remember to update `status_since` to current time in UTC.

Auch-Auch:
Closes #267

### DIFF
--- a/api/processors/actions/__init__.py
+++ b/api/processors/actions/__init__.py
@@ -18,7 +18,6 @@ __all__ = [
     "StudentTransferProcessor",
     "StudentMissedClassProcessor",
     "TeacherTransferProcessor",
-    "StudentWantOnLeave",
     "TeacherWentOnLeaveProcessor",
     "StudentWentOnLeaveProcessor",
 ]
--- a/api/processors/actions/__init__.py
+++ b/api/processors/actions/__init__.py
@@ -5,7 +5,9 @@ from api.processors.actions.group.finish import GroupFinishProcessor
 from api.processors.actions.group.start import GroupStartProcessor
 from api.processors.actions.student.missed_class import StudentMissedClassProcessor
 from api.processors.actions.student.transfer import StudentTransferProcessor
+from api.processors.actions.student.went_on_leave import StudentWentOnLeaveProcessor
 from api.processors.actions.teacher.transfer import TeacherTransferProcessor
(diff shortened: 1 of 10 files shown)


api/processors/actions/student/went_on_leave.py
```diff
@@ -0,0 +1,27 @@
...
+        self._set_statuses()
+
+    def _set_statuses(self) -> None:
+        self.student.project_status = StudentProjectStatus.ON_LEAVE
+        self.student.status_since = self.timestamp
+        self.student.save()
+
+    def _create_log_events(self) -> None:
+        StudentLogEvent.objects.create(
+            student=self.student, type=StudentLogEventType.GONE_ON_LEAVE
```
Auch-Auch, line 26: Note that group is None

lemontree210: I think it makes sense. The student will either drop out of the group first or (unlikely, though?) just continue with the same group when they return.

api/processors/actions/__init__.py
```diff
@@ -16,4 +18,7 @@
     "StudentTransferProcessor",
     "StudentMissedClassProcessor",
     "TeacherTransferProcessor",
+    "StudentWantOnLeave",
```
lemontree210, line 21:
```suggestion
    "StudentWentOnLeave",
```
or maybe something completely different should be here (or nothing?)

lemontree210 (approved):
Thank you! Just one typo, I think.

api/processors/actions/__init__.py
Auch-Auch: Ah, it is just redundant default import, it is not needed here

PR #277, opened 2023-09-27
Student and teacher returned from leave view

### LINKED ISSUE #275: API view: `Student` or `Teacher` returned from leave
(cf. #267 that was fixed by #273)

When a coordinator from dashboard or bot sends a request stating that a student or a teacher returned from leave:

## Create log events

Type of `LogEvent`s to be created: `RETURNED_FROM_LEAVE`

## Set statuses

Depending on whether the number of groups is `> 0`

- `Student`: `NO_GROUP_YET` or `STUDYING`
- `Teacher`: `NO_GROUP_YET` or `WORKING`

> [!Important]
> Remember to update `status_since` to current time in UTC.

Auch-Auch:
Closes #275


api/models/teacher.py
```diff
@@ -115,3 +115,7 @@ class Meta:
     def can_take_more_groups(self) -> bool:
         """`True` if a teacher can take more groups than they already have."""
         return self.groups.count() < self.simultaneous_groups
+
+    @property
+    def has_groups(self) -> bool:
+        return self.groups.count() > 0
```
Auch-Auch, line 121: Note that I count all groups but maybe we should count only active groups? (same for student model)

Auch-Auch:
Just noticed that both returned_from_leave and went_on_leave should be in internal router as well :)

lemontree210:
> Just noticed that both returned_from_leave and went_on_leave should be in internal router as well :)

should they? I don't think we plan to have them in the bot yet. In the issues, I intentionally only wrote about coordinators performing the actions in the dashboard.

lemontree210:
> > When a coordinator from dashboard or bot sends a request stating that a student or a teacher returned from leave:
> 
> Coordinator or bot

Dammit! 😆 Well, I can imagine that in the future student/teacher will be able to tell the bot "I'm unavailable until..." which the bot will use to set this status.

api/migrations/0001_initial.py
```diff
@@ -568,6 +572,8 @@ class Migration(migrations.Migration):
                             ("left", "Left the project"),
                             ("expelled", "Expelled from the project"),
                             ("access_revoked", "Access to corporate resources revoked"),
+                            ("gone_on_leave", "Gone on leave"),
+                            ("returned_from_leave", "Returned from leave"),
```
lemontree210, line 576: we forgot to recreate the 1st migration in #273, right?

api/processors/actions/student/returned_from_leave.py
```diff
@@ -0,0 +1,30 @@
...
+    @transaction.atomic
+    def process(self) -> None:
+        self._create_log_events()
+        self._set_statuses()
+
+    def _set_statuses(self) -> None:
+        if self.student.has_groups:
+            self.student.project_status = StudentProjectStatus.STUDYING
+        else:
+            self.student.project_status = StudentProjectStatus.NO_GROUP_YET
```
lemontree210, line 23: nit: could be a ternary operator, but sure it'll fit in one line though

tests/tests_api/test_student.py
```diff
@@ -499,11 +506,74 @@ def test_dashboard_student_went_on_leave(api_client, availability_slots, timesta
...
+            "/api/dashboard/students/",
+            "/api/students/",
+        ],
+    )
+    def test_no_group(self, api_client, availability_slots, timestamp, endpoint):
+        student = baker.make(
+            Student,
+            make_m2m=True,
+            _fill_optional=True,
+            project_status=StudentProjectStatus.LEFT_PREMATURELY,
```
lemontree210, line 532: I know this is irrelevant right now, but I'd rather either use some different status (because `LEFT_PREMATURELY` is for the situation when the person already left the project, so they wouldn't go on leave) or let faker choose some random status. The reason why I don't want this `LEFT_PREMATURELY` here is that if we introduce some logic for checking status changes, this test here will break for no apparent reason.

api/views/student.py
```diff
@@ -17,10 +17,19 @@
...
+)
 
 
-class StudentViewSet(ReadWriteSerializersMixin, viewsets.ModelViewSet[Student]):  # type: ignore
+class StudentViewSet(  # type: ignore
+    ReadWriteSerializersMixin,
+    viewsets.ModelViewSet[Student],
+    StudentWentOnLeaveMixin,
+    StudentReturnedFromLeaveMixin,
+):
```
lemontree210, line 32: Hm... Why is it so that after so many actions we have already added (e.g. transfer from group to group etc.) this is the first time we're adding some mixins to the viewsets?

lemontree210 (commented):
I guess the only really substantial question here is about the addition of mixins to viewsets

api/models/student.py
```diff
@@ -67,3 +67,7 @@ class Meta:
             models.Index(fields=("project_status",), name="student_pr_status_idx"),
             models.Index(fields=("situational_status",), name="student_si_status_idx"),
         ]
+
+    @property
+    def has_groups(self) -> bool:
+        return self.groups.count() > 0
```
Roxe322, line 73: Maybe `self.groups.exists()` ?
https://dev.to/codereviewdoctor/why-queryset-exists-is-more-efficient-than-queryset-count-2f3h

api/models/teacher.py
Roxe322: Same as above

api/migrations/0001_initial.py
Auch-Auch: Yep

api/views/student.py
Auch-Auch: Because `Coordinator or bot` were only in the last two issues :)

tests/tests_api/test_student.py
Auch-Auch: Okay, removed it

lemontree210 (approved):
Thank you!

PR #280, opened 2023-10-01
Add available students view

### LINKED ISSUE #276: API view: show (number of) students available for given day/time slots
After some consultation with UX we realized that the following workflow will be optimal for coordinators when creating a tentative group in the dashboard:

1. Coordinator finds a teacher and selects him/her. Coordinator chooses language, age group(s), level and communication language.
2. Then, there are some pairs of slots that the coordinator wants to explore to find out which pair is best for creating a tentative group.
3. When the coordinator chooses a pair of slots, the dashboard will send a request to the backend, and the backend must return **the number of students** (with matching parameters and project_status `NO_GROUP_YET`), so that the coordinator can choose the most promising option that will fit as many students as possible.
4. This API view can return the IDs and names of the students as well (IDs for sending back to the backend when the tentative group is created, names for displaying in the form on the next stage). The question is, however, whether ToolJet can store the data (I think that only the data from the latest API request will need to be saved in some sort of variable) or whether another API request is needed when the coordinator chooses the pair of slots. If IDs and names are returned, maybe the number is not needed because the size of the list can be calculated in ToolJet.

Auch-Auch:
Closes #276


api/views/student.py
```diff
@@ -105,6 +109,34 @@ def missed_class(self, request: Request, personal_info_id: int) -> Response:  #
...
+            ),
+        },
+    )
+    @action(detail=False, methods=["get"])
+    def available_students_list(self, request: Request) -> Response:  # noqa: ARG002
+        query_params_serializer = DashboardAvailableStudentsSerializer(
+            data=request.query_params,
+        )
+        query_params_serializer.is_valid(raise_exception=True)
+        matched_students = Student.objects.filter(
```
Auch-Auch, line 131: I know it is bad to have it here but I though that creating a separate processor might be too much, if you have ideas to where exactly put it feel free to tell

api/views/student.py
```diff
@@ -105,6 +109,38 @@ def missed_class(self, request: Request, personal_info_id: int) -> Response:  #
...
+                description="Students returned",
+            ),
+            status.HTTP_400_BAD_REQUEST: OpenApiResponse(
+                response=ValidationErrorSerializer,
+                description="Something is wrong with the query params",
+            ),
+            status.HTTP_409_CONFLICT: OpenApiResponse(
+                response=BaseAPIExceptionSerializer,
+                description="Invalid time slots",
+            ),
```
lemontree210, line 126: I think that `400 BAD REQUEST` should be used for both cases, we should give exact reason in the response body. If the dashboard gives wrong IDs or only one slot instead of two, it's definitely a `BAD REQUEST` IMO. `409 CONFLICT` could be returned when no students are found, but IIRC we voted and decided that `200 OK` be returned with an empty list in such cases.

lemontree210 (changes requested):
Thank you for keeping up the great work 💪🏻

api/views/student.py
Auch-Auch: Here probably the message is confusing, but I can change the message to "some of the time slots not found" or change the status as well. I used 409 everywhere where some of the body items were not found but here we have query param so it might be reasonable to use `not found`

lemontree210: Okay, according to [this vote](https://t.me/c/1808422267/1258), in a similar case we decided to return 200 OK and empty list, so let's do the same here.

api/views/student.py
```diff
@@ -105,6 +109,38 @@ def missed_class(self, request: Request, personal_info_id: int) -> Response:  #
...
+                description="Students returned",
+            ),
+            status.HTTP_400_BAD_REQUEST: OpenApiResponse(
+                response=ValidationErrorSerializer,
+                description="Something is wrong with the query params",
+            ),
+            status.HTTP_404_NOT_FOUND: OpenApiResponse(
+                response=BaseAPIExceptionSerializer,
+                description="Some of time slots not found",
+            ),
```
lemontree210, line 126: I guess I still don't understand what it means that "some of the time slots [were] not found" 😕 Does it mean that IDs of time slots are wrong? In this case it should be 400 bad request as well. I don't think this case is covered by tests so it's hard to tell exactly what I'm not getting :)

api/views/student.py
lemontree210: I see that throughout our code `409` is used for `POST` and `DELETE` requests, and I think this is consistent with recommended practice. So it was the right call to get rid, but I'm not sure about `404`. I guess we'll continue this discussion under [the newer comment](https://github.com/samanthasgroup/django-webapps/pull/280#discussion_r1349576910).

api/views/student.py
lemontree210: How about status code [422 Unprocessable Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422)? Seems fitting:

> The server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.

lemontree210: I don't think this case is covered by tests BTW.

Auch-Auch: Okay then

lemontree210 (approved):
Thank you!

PR #295, opened 2023-10-10
Add views to be able to control students next steps after studies

### LINKED ISSUE #292: API view: determine next step for `Student` after group finished studies
When a group finishes studies (#234, fixed by #235), the students that were in this group still retain `STUDYING` status (see [discussion](https://github.com/samanthasgroup/django-webapps/pull/235/files#r1306648977) in #235, where we decided not to create a "limbo" status). Hence, please delete this `project_status`: `AWAITING_DECISION = "awaiting_decision", "Awaiting decision"`

After this, we need to:

- show to coordinators in TJ dashboard the list of students that are awaiting decision
Maybe we also have to show some mark in the student's profile. Anyway, maybe this can be done without an additional view (a piece of code in TJ could check if the student has status `STUDYING` and at the same time no groups). @dont-text-me what do you think?
- in student's profile, add functionality for the coordinator to choose the next action: either to mark the student as leaving the project completely, or put them back into the queue (maybe with an updated language level)

So the tasks regarding API views would be the following (a separate issue can be created from each task, just click on a circle right of the task)
- [ ] API view: return students with status `STUDYING` but no groups
- [ ] API view: mark student as leaving the project completely by creating log event of type `FINISHED_AND_LEAVING` and setting `project_status` to `FINISHED`.
- [ ] API view: return the student to the waiting queue by creating log event of type `AWAITING_OFFER` and setting `project_status` to `NO_GROUP_YET`.

Auch-Auch:
Closes #292


.pre-commit-config.yaml
```diff
@@ -26,6 +26,8 @@ repos:
             - "django_extensions"
             - "drf-spectacular"
             - "django-filter"
+        args:
+            - --no-incremental
```
Auch-Auch, line 30: Note: it will disable cache, without it mypy is no longer working :(

api/processors/actions/group/finish.py
```diff
@@ -69,7 +69,7 @@ def _set_students_status(self) -> None:
 
         # after moving the student, this 1 will become 0
         annotated_students.filter(groups_count=1).update(
-            project_status=StudentProjectStatus.AWAITING_DECISION,
+            project_status=StudentProjectStatus.NO_GROUP_YET,
```
Auch-Auch, line 72: Assumed `NO_GROUP_YET' is what we need here

api/processors/student.py
```diff
@@ -26,3 +28,11 @@ def went_on_leave(student: Student) -> None:
     @staticmethod
     def returned_from_leave(student: Student) -> None:
         StudentReturnedFromLeaveProcessor(student).process()
+
+    @staticmethod
+    def finished_and_left(student: Student) -> None:
```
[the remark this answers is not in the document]
lemontree210: Nah, we have `missed_class` or something, this is similar

api/views/student.py
```diff
@@ -109,6 +109,42 @@ def missed_class(self, request: Request, personal_info_id: int) -> Response:  #
...
+            ),
+        },
+    )
+    @action(detail=True, methods=["post"])
+    def finished_and_left(
+        self, request: Request, personal_info_id: int  # noqa: ARG002
+    ) -> Response:
+        student = self.get_object()
+        if student.has_groups:
+            raise ConnectionError("Only students with no groups can be processed")
```
lemontree210, line 127: just checking: is `ConnectionError` really an appropriate exception here?

lemontree210 (changes requested):
I would rename `put_on_wait` (and respective class name) to something like `put_back_in_waiting_queue`. It's just that I saw the name and for a minute couldn't understand what it was referring to :)

api/views/student.py
Auch-Auch: Good you noticed, it is wrong

api/views/student.py
```diff
@@ -141,6 +177,24 @@ def available_students_list(self, request: Request) -> Response:  # noqa: ARG002
...
+            status.HTTP_200_OK: OpenApiResponse(
+                response=ListSerializer(child=DashboardStudentSerializer()),
+                description="Students returned",
+            ),
+        }
+    )
+    @action(detail=False, methods=["get"])
+    def active_students_with_no_groups(self, request: Request) -> Response:  # noqa: ARG002
+        matched_students = Student.objects.filter(
+            project_status=StudentProjectStatus.STUDYING
```
Roxe322, line 191: Don't we need any additional filter by situational status here? @lemontree210

Roxe322 (approved):

api/views/student.py
lemontree210: Thanks for checking - no, I don't think so

Auch-Auch:
PR is broken somehow

PR #303, opened 2023-10-13
Add offer join group view

Auch-Auch:
Closes #298

PR #304, opened 2023-10-14
Add accepted offered group view and tests

### LINKED ISSUE #299: API view: add student that accepted the offer to the existing group
This is a follow-up on #298 

When a coordinator from dashboard sends a request stating that a student that was earlier offered to join a group has accepted this offer:

## Add new log event type to `CoordinatorLogEventType`:

```py
    ADDED_STUDENT_TO_EXISTING_GROUP = (
        "added_student_to_existing group",
        "Added student to a group that was already studying",
    )
```

## Create log events

- coordinator(s): `ADDED_STUDENT_TO_EXISTING_GROUP`
- student: `STUDY_START`

## Set statuses
- student: set `project_status` to `STUDYING`, `situational_status` to `""` (blank)

> [!Note]
> Remember about `status_since` :)

Auch-Auch:
Closes #299


api/processors/actions/student/accepted_offered_group.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def _set_statuses(self) -> None:
+        self.student.project_status = StudentProjectStatus.STUDYING
+        self.student.situational_status = ""
+        self.student.status_since = self.timestamp
+        self.student.save()
+
+    def _create_log_events(self) -> None:
+        StudentLogEvent.objects.create(
+            student=self.student, to_group=self.group, type=StudentLogEventType.STUDY_START
```
Auch-Auch, line 31: Please, double check that, we also have event `accepted_offer`

Auch-Auch:
Also noticed that we have tons of status setters already, maybe at some point it would be nice to double check them before the realease

Roxe322 (approved):

lemontree210:
@Auch-Auch please merge `master` with conflict resolution (conflict emerged when I merged #303 into `master`)

api/processors/actions/student/accepted_offered_group.py
lemontree210: Yes, I know, but thanks for checking. I think this event will be used with the automatic communication with students when a new group is created. Maybe at some point we will introduce the functionality when a coordinator clicks something like "propose to join existing group" and the bot contacts the student. Then this `accepted_offer` event may be needed. For now, though, the situation is that the group is already studying and the coordinator contacts another student and asks them: do you want to join group such-and-such? If the student says yes, the coordinator adds them to the group and the student is already considered "studying". Hence `STUDY_START`. Of course, we could create two events simultaneously, but I don't see any value in that.

PR #305, opened 2023-10-15
Added groups, former groups and logevents to CoordinatorAdmin

### LINKED ISSUE #296: Admin site: Show this coordinator's groups
- Show number of active groups in for each coordinator in list of coordinators
- Show all groups of a given coordinator (`.groups`, after them `.groups_former`, sorted by creation date or start date) in this coordinator's profile. The information about each groups can be concise. `Group`'s `__str__()` method gives enough information, but we can use this issue to add the group's status to its `__str__()` and also fix missing closing parenthesis :)

### LINKED ISSUE #297: Admin site: Show this coordinator's log events in coordinator's profile
Not sure if any description is needed here :) Let's do basic functionality for now, without any hyperlinks to pages of groups etc.

> [!important]
> Some log events require admin's ability to choose `group`:
    `TOOK_NEW_GROUP`
    `GROUP_STARTED_CLASSES`
    `TOOK_TRANSFERRED_GROUP`
    `REQUESTED_TRANSFER`
    `TRANSFER_CANCELED`
    `TRANSFER_COMPLETED`
    `GROUP_ABORTED`
    `GROUP_FINISHED`

Roxe322:
Split admin.py to separate modules

Closes #296 
Closes #297


lemontree210 (changes requested):
Unfortunately, the solution with choosing an action from drop-down list and pressing "Go" doesn't seem to work in cases when we need to indicate a `.group`. A log event is created with empty `.group` because they can't be chosen.

api/processors/auxil/log_event_creator.py
```diff
@@ -139,11 +139,13 @@ def create(
         coordinator: Coordinator,
         log_event_type: CoordinatorLogEventType,
         comment: str = "",
+        group: Group | None = None,
     ) -> None:
         created_log_event = CoordinatorLogEvent.objects.create(
             coordinator=coordinator,
             type=log_event_type,
             comment=comment,
+            group=group,
```
lemontree210, line 148: Ideally we should check if the given event requires a `.group`, or, on the contrary, must not have any `.group` so that the user cannot create a wrong log event. What do you think?

lemontree210 (approved):
Thank you!

PR #322, opened 2023-10-25
Add view to confirm that student finished oral interview

### LINKED ISSUE #320: API view: oral interview successful, can wait for group
Some students (e.g. those that can't read or haven't learned English long enough) require a one-to-one interview. The student's language level is determined in this interview, then they can be put in the waiting queue.

A coordinator send from dashboard a request stating that the given `Student` has a given language level and can be put into the queue. The data in a `POST` request must include ID of `LanguageAndLevel` object.

## Check `Student`'s `project_status`

If the given `Student`'s `project_status` is not `NEEDS_INTERVIEW_TO_DETERMINE_LEVEL`, this is a wrong request.

## Change objects

Create a new many-to-many connection between `Student` and `LanguageAndLevel` by adding the `LanguageAndLevel` object to `.teaching_languages_and_levels`.

> [!NOTE] 
> Most likely all students will only be allowed to study in one group, but the database was designed to be more flexible and to technically allow several groups.

## Set status

Set `project_status` to `NO_GROUP_YET`

## Create log events

Add log event of type `AWAITING_OFFER`
(**TODO:** check what event type is created when student is created through bot)

Auch-Auch:
Closes #320

### DIFF
--- a/api/views/student.py
+++ b/api/views/student.py
@@ -13,7 +13,7 @@ from api.models.choices.status.project import StudentProjectStatus
 from api.processors import StudentProcessor
 from api.serializers import (
     DashboardAvailableStudentsSerializer,
-    DashboardFinishedOralInterviewSerializer,
+    DashboardCompletedOralInterviewSerializer,
     DashboardStudentAcceptedOfferedGroupSerializer,
     DashboardStudentMissedClassSerializer,
     DashboardStudentOfferJoinGroupSerializer,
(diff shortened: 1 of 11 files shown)


api/views/student.py
```diff
@@ -293,6 +294,36 @@ def active_students_with_no_groups(self, request: Request) -> Response:  # noqa:
...
 
+    @extend_schema(
+        request=DashboardFinishedOralInterviewSerializer,
+        responses={
+            status.HTTP_204_NO_CONTENT: OpenApiResponse(description="Action is taken"),
+            status.HTTP_409_CONFLICT: OpenApiResponse(
+                response=BaseAPIExceptionSerializer,
+                description="Only with status NEEDS_INTERVIEW_TO_DETERMINE_LEVEL can be processed",
+            ),
+            status.HTTP_422_UNPROCESSABLE_ENTITY: OpenApiResponse(
```
Auch-Auch, line 305: I suggest using this status everywhere where we have not found some entity, so use it instead of 409, who do you think?

lemontree210: yes, let's go for 422. I searched through old messages in our TG chat and see that @Roxe322 was in favor of 422 over 409 in similar cases. The main argument for 422 is that it means that the client [should not repeat the request without modificatio](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422), which is exactly what we want to tell the client. 409 doesn't necessary entail that.

lemontree210 (changes requested):
I would ask you to replace "finished" with e.g. "completed" (in urls and class names). Finished can be confused with "finished studies".

I could also suggest "passed" but you can't actually fail this interview.

lemontree210 (approved):
LGTM. Thanks for being so thorough :)

PR #329, opened 2023-10-28
Add log events on person creation

### LINKED ISSUE #323: Create log events when new coordinator, student or teacher are created
Right now instances of `Coordinator`, `Student` and `Teacher` are created with the correct `project_status`, but we also need to create log events.

- `Coordinator`: `APPLIED`
- `Students`: `REGISTERED`
- `Teacher`: `REGISTERED`

It's OK to create two log events at the same time, e.g. for a `Student` that has a level determined and doesn't need any oral interview, we can create two log events right away: `REGISTERED` and `AWAITING_OFFER` (same for `Teacher`). It still makes sense especially for `Student` because for students that need oral interview, log event of type `REGISTERED` will not be immediately followed by log event of type `AWAITING_OFFER`.

Also, please check that `Group` gets a log event of type `FORMED` when it's created.

Auch-Auch:
Closes #323

### DIFF
--- a/api/serializers/student/internal.py
+++ b/api/serializers/student/internal.py
@@ -1,6 +1,10 @@
+from typing import Any
+
 from rest_framework import serializers
 
+from api.exceptions import ConflictError
 from api.models import LanguageAndLevel, Student
+from api.models.choices.status.project import StudentProjectStatus
 from api.serializers import (
     AgeRangeSerializer,
     DayAndTimeSlotSerializer,
@@ -12,6 +16,19 @@ class StudentWriteSerializer(serializers.ModelSerializer[Student]):
     teaching_languages_and_levels = serializers.PrimaryKeyRelatedField(
         queryset=LanguageAndLevel.objects.all(), required=False, allow_empty=True, many=True
     )
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        if (
+            not attrs["teaching_languages_and_levels"]
+            and attrs["project_status"]
+            != StudentProjectStatus.NEEDS_INTERVIEW_TO_DETERMINE_LEVEL.value
+        ):
+            raise ConflictError(
+                """Poject status should be needs_interview_to_determine_level 
+                   if language and level is not specified"""
+            )
+
+        return attrs
 
     class Meta:
         model = Student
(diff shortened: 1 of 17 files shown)


api/serializers/student/internal.py
```diff
@@ -9,6 +9,10 @@
 
 
 class StudentWriteSerializer(serializers.ModelSerializer[Student]):
+    teaching_languages_and_levels = serializers.PrimaryKeyRelatedField(
+        queryset=LanguageAndLevel.objects.all(), required=False, allow_empty=True, many=True
```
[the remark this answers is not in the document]
lemontree210: That's a tricky one. On the one hand we do, because there are students that get no level, but on the other hand we should make sure language and level is not omitted by mistake. I guess that, ideally, we should make it optional but throw an exception if level is empty but `project_status` is not `NEEDS_INTERVIEW_TO_DETERMINE_LEVEL`

tests/tests_api/test_student.py
```diff
@@ -132,7 +149,7 @@ def test_student_update(api_client, availability_slots):
     )
     Student.objects.get(pk=student.pk)
     fields_to_update = {
-        "project_status": StudentProjectStatus.BANNED,
+        "project_status": StudentProjectStatus.NEEDS_INTERVIEW_TO_DETERMINE_LEVEL,
```
Auch-Auch, line 152: One thing that we need to remember is that if we update student project_status and language and level are empty then it will throw an error. So we kind of can proceed with a student only if it has language and level

api/serializers/student/internal.py
```diff
@@ -13,6 +17,28 @@ class StudentWriteSerializer(serializers.ModelSerializer[Student]):
...
+            levels = attrs.get("teaching_languages_and_levels")
+        if (
+            project_status
+            and not levels
+            and project_status != StudentProjectStatus.NEEDS_INTERVIEW_TO_DETERMINE_LEVEL.value
+        ):
+            raise ConflictError(
+                """Poject status should be only needs_interview_to_determine_level 
+                   if language and level is not specified"""
+            )
```
lemontree210, line 38:
```suggestion
            msg = (
                f'Project status can only be "{StudentProjectStatus.NEEDS_INTERVIEW_TO_DETERMINE_LEVEL.value}"'
                "if language and level is not specified"
            )
            raise ConflictError(msg)
```

lemontree210 (approved):
just one cosmetic change and it LGTM, so I'm approving before even you make it

tests/tests_api/test_student.py
lemontree210: It is, yes. After the interview, the coordinator must edit the student's record in the dashboard and set their level

PR #330, opened 2023-10-30
Remove redundant save() calls

### LINKED ISSUE #310: refactor: check for redundant `.save()` calls
There are some `.save()` calls when objects are removed from `.groups()` or `.students()` and the like (m2m relationships) and no other changes (such as `.status_since`) are made.

AFAIK, we don't need to call `.save()` after `.add()`, `.remove()` or `.clear()` in m2m relationships because these operations affect intermediary tables, and the row of the table with objects is not changed.

Roxe322:
Closes #310


lemontree210 (approved):
LGTM

PR #336, opened 2023-11-01
Add teacher finished studies and stays in project view

### LINKED ISSUE #333: API view: finished teaching but stays in the project
When a coordinator (from dashboard) sends a request stating that `Teacher` finished teaching but is staying in the project:

## Check `Teacher`'s groups

Make sure the teacher has no `.groups()`. They have to be finished properly (or teacher has to be removed from them) before the teacher can be marked as "finished teaching". Return status code `422` in case of error (**UPD** `409 CONFLICT` may be more appropriate, see https://github.com/samanthasgroup/django-webapps/pull/336#discussion_r1380854537)

## Create log events

Type of `LogEvent` to be created: `FINISHED_AND_STAYING`

## Set statuses

- `project_status` to `FINISHED_STAYS`
- `situational_status` to `""`

> [!note]

Maybe the URL route should be something like `finished_but_stays_in_project` (with the word "project")

Auch-Auch:
Closes #333

### DIFF
--- a/api/processors/actions/teacher/finished_but_stays_in_project.py
+++ b/api/processors/actions/teacher/finished_but_stays_in_project.py
@@ -1,0 +1,33 @@
+from django.db import transaction
+
+from api.models import Teacher
+from api.models.choices.log_event_type import TeacherLogEventType
+from api.models.choices.status.project import TeacherProjectStatus
+from api.models.log_event import TeacherLogEvent
+from api.processors.actions.teacher import TeacherActionProcessor
+
+
+class TeacherFinishedStudiesButStaysInProjectProcessor(TeacherActionProcessor):
+    def __init__(self, teacher: Teacher):
+        super().__init__(teacher)
+
+    @transaction.atomic
+    def process(self) -> None:
+        self._update_groups()
+        self._create_log_events()
+        self._set_statuses()
+
+    def _update_groups(self) -> None:
+        if self.teacher.has_groups:
+            self.teacher.groups_former.add(*self.teacher.groups.all())
+            self.teacher.groups.clear()
+
+    def _set_statuses(self) -> None:
+        self.teacher.project_status = TeacherProjectStatus.FINISHED_STAYS
+        self.teacher.status_since = self.timestamp
+        self.teacher.save()
+
+    def _create_log_events(self) -> None:
+        TeacherLogEvent.objects.create(
+            teacher=self.teacher, type=TeacherLogEventType.FINISHED_AND_STAYING
+        )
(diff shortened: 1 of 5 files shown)


api/processors/actions/teacher/finished_but_stays_in_project.py
```diff
@@ -0,0 +1,33 @@
...
+        super().__init__(teacher)
+
+    @transaction.atomic
+    def process(self) -> None:
+        self._update_groups()
+        self._create_log_events()
+        self._set_statuses()
+
+    def _update_groups(self) -> None:
+        if self.teacher.has_groups:
```
Auch-Auch, line 21: This will never be executed since we check it in the view but I still implemented it in case we will use it separately. Another option is to throw an exception here

api/views/teacher.py
```diff
@@ -123,6 +124,29 @@ def expelled(  # noqa: ARG002
...
+    @extend_schema(
+        responses={
+            status.HTTP_204_NO_CONTENT: OpenApiResponse(description="Action is taken"),
+            status.HTTP_400_BAD_REQUEST: OpenApiResponse(
+                response=ValidationErrorSerializer,
+                description="Something is wrong with the query params",
+            ),
+            status.HTTP_422_UNPROCESSABLE_ENTITY: OpenApiResponse(
+                response=BaseAPIExceptionSerializer,
+                description="Unable to process teacher with group(s)",
```
Auch-Auch, line 136: Actually, a ConlifError seems good to me here too

api/processors/actions/teacher/finished_but_stays_in_project.py
lemontree210: I would throw an exception. If there are groups here, it is a coding error here in django_webapps, not user error (user = dashboard developer that programmed the request). As for which exception, I'm not sure. Maybe `ValueError`

api/views/teacher.py
```diff
@@ -63,15 +64,15 @@ class DashboardTeacherViewSet(
             status.HTTP_204_NO_CONTENT: OpenApiResponse(description="Teacher is transferred"),
             status.HTTP_409_CONFLICT: OpenApiResponse(
                 response=BaseAPIExceptionSerializer,
-                description="Transfer group is not found",
+                description="Teacher is already in transfer_to group or not in transfer_from",
```
lemontree210, line 67:
```suggestion
                description='Teacher is not in "from_group" or already in "to_group"',
```

api/views/teacher.py
lemontree210: > Actually, a ConlifError seems good to me here too

Maybe you're right. `422` signalizes that the client should not send the request with the same data again. But actually, if the client removes the groups in some other request, then they can actually resend the request here with the same teacher data - and succeed.

lemontree210 (commented):
I'm impressed by your rhythm, great job 👏🏻

lemontree210 (approved):
LGTM

Roxe322 (approved):

PR #342, opened 2023-11-04
Added django-reversion integration

### LINKED ISSUE #332: Add version control for model instances
In our internal discussion, taking into consideration the active maintainable status of the project, we decided to use [django-reversion](https://github.com/etianen/django-reversion).

Ideally, even when changes are made to the model instances via API (from dashboard and bot), new revision should be created, if possible. But at least for those that are created in the admin interface.

@dont-text-me I don't think that we need to implement the functionality for reverting changes in the dashboard, at least not in the MVP. For now, it will be only available to the admins in the admin site.

Roxe322:
Closes #332

(!) Don't forget to update local settings.py and run `python manage.py createinitialrevisions`


lemontree210 (commented):
maybe we should add this `python manage.py createrevisions` to some commands in our `Makefile`?

Roxe322:
@lemontree210 I don't think so. `createrevisions` has to be called only when new model is registered in admin, and once on server inititalization. So I didn't find any suitable commands in Makefile to include it.

lemontree210 (approved):

PR #343, opened 2023-11-05
Feature/teacher access revoked

### LINKED ISSUE #334: API view: access (to teachers' resources) revoked; finished and leaving
When a coordinator (from dashboard) sends a request stating that `Teacher` finished teaching, their access to teaching resources was revoked and they left the project:

## Check `Teacher`'s groups

Make sure the teacher has no `.groups()`. They have to be finished properly (or teacher has to be removed from them) before the teacher can be marked as "finished teaching". Return status code `422` in case of error.

> [!IMPORTANT]
> We **do not** send any requests to terminate teacher's access to external resources or check that it was terminated, we trust the coordinator to do it manually before sending the request to the backend from the dashboard. Maybe someday we will introduce some functionality that will make sure the access was really revoked, but so far we don't even have such plans.

## Create log events

Type of `LogEvent` to be created: `ACCESS_REVOKED`, then `FINISHED_AND_LEAVING`. 
I guess that the timestamp for `ACCESS_REVOKED` could be set to, say, one second before the common timestamp for the second log event and for `status_since`.

> [!NOTE]
> For future thinking: we don't create an `ACCESS_REVOKED` log event when teacher is expelled, but then it's sort of self-explanatory that this must be done when such incident happens. In case of "finished and left", there can be a significant amount of time that passes between end of teaching and the moment of leaving the project, so maybe it still make sense to have an explicit log event.

## Set statuses

- `project_status` to `FINISHED_LEFT`
- `situational_status` to `""`

> [!note]

Auch-Auch:
Closes #334


api/models/log_event.py
```diff
@@ -29,7 +30,7 @@ class LogEvent(models.Model):
     """
 
     comment = models.TextField()
-    date_time = models.DateTimeField(auto_now_add=True)
+    date_time = models.DateTimeField(default=timezone.now)
```
Auch-Auch, line 33: Note: https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.DateField.auto_now_add

lemontree210: I see, yes. Since we now want to modify `date_time`, we do have to make this change 👍🏻

lemontree210 (changes requested):
Thank you for the PR and for taking care of the auto datetime.

I'd rather make this view focused on the "finished and leaving" event rather than "revoked". One reason is a balance with the "finished and stays" view, and the other is that we may create a separate "access revoked" view in the future, if e.g. we have some integration between backend and Notion.

So I'd ask to change the names of the processor and the action view.

lemontree210 (approved):
Thank you!

PR #366, opened 2023-12-07
feat(admin site): change `PersonalInfo.__str__()`, use it in other models

### LINKED ISSUE #361: Models: change output of `__str__` method to `{id}: {full_name}`, use it in other models
- adjust `PersonalInfo__str__()` for it to produce `f"{self.id}: {self.full_name}"`
- in `Coordinator`, `Student` and `Teacher...`, set `__str__()` to return `str(self.personal_info)`
- in admin site for `Group` (#351), replace manually constructed representation (ID, colon, full name) with the output of `__str__` 

This will give us the uniform representation everywhere and will help reduce repetitions in future admin views.

> [!NOTE]
> You may need to run the Makefile script for recreating first migration when you change the models' `__str__()` methods, I'm not sure though that it's included in the migration file

Ross1832:
closes #361

### DIFF
--- a/api/models/personal_info.py
+++ b/api/models/personal_info.py
@@ -88,10 +88,9 @@ class PersonalInfo(GroupOrPerson):
             update_fields=update_fields,
         )
 
+    def __str__(self) -> str:
+        return f"{self.pk}: {self.full_name}"
+
     @property
     def full_name(self) -> str:
         return f"{self.first_name} {self.last_name}"
-
-    @property
-    def pk_full_name(self) -> str:
-        return f"{self.pk}: {self.full_name}"
--- a/api/models/personal_info.py
+++ b/api/models/personal_info.py
@@ -71,9 +71,6 @@ class PersonalInfo(GroupOrPerson):
         ordering = ("last_name", "first_name")
         verbose_name_plural = "personal info records"
 
-    def __str__(self) -> str:
-        return f"{self.full_name} ({self.pk})"
-
     def save(
         self,
         force_insert: bool = False,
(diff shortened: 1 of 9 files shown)


lemontree210 (changes requested):
I think that you should not have deleted `PersonalInfo.__str__()`.

Instead of creating the new property `.pk_full_name()` and then calling it, e.g. in `admin/group.py`, you could do:

```py
def __str__(self) -> str:
        return f"{self.pk}: {self.full_name}"
```
then, in group.py, you could just put

```py
format_html(
    ...
    ...,
    teacher
)
```
and that's it. As far as I understand `teacher.__str__()` will be called, which in turn will call `teacher.personal_info.__str__()`, which will return "123: John Smith".

Or am I missing something?
```

File: api/models/personal_info.py [hunk 1]
```diff
@@ -94,3 +91,7 @@ def save(
     @property
     def full_name(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+    @property
+    def pk_full_name(self) -> str:
+        return f"{self.pk}: {self.full_name}"
```
lemontree210, line 97:
```suggestion
    def __str__(self):
        return f"{self.personal_info}"
```

lemontree210 (changes requested):

Ross1832:
I have applied changes suggested in the feedback comments

lemontree210:
> I have applied changes suggested in the feedback comments

Thank you! Perhaps I didn't explain well enough what I had in mind :)
I implemented this in #31ee964

lemontree210 (approved):
LGTM

Roxe322 (approved):

PR #402, opened 2024-01-21
#403 Group builder priority by levels and students' waiting time

### LINKED ISSUE #403: Add logic for student prioritisation in group building (given level, language, time slots, and waiting time)

Roxe322:
Closes #403


api/processors/services/group_builder.py
```diff
@@ -51,7 +54,7 @@ class GroupSizeRestriction:
     def validate_size(self, size: int) -> None:
         if self.min > size:
             raise ValueError(
-                f"Group candidate must respect upper group boundaries, but {size} > {self.max}"
+                f"Group candidate must respect upper group boundaries, but {size} < {self.min}"
```
lemontree210, line 57: it says "upper", so the code made sense. I remember discussing it with @paulin-mipt , but I don't remember how she managed to explain it to us why it actually made sense :) Too bad there's no comment

api/processors/services/group_builder.py
```diff
@@ -282,3 +276,37 @@ def _create_log_events(group: Group) -> None:
...
+        level_difference = abs(
+            group1.language_and_level.level.index - group2.language_and_level.level.index
+        )
+        time_threshold = timezone.now() - MAX_WAITING_TIME
+
+        if level_difference >= LEVEL_DIFFERENCE_THRESHOLD:
+            # If levels difference is 2 or more, prefer the lower level
+            if group1.language_and_level.level.index < group2.language_and_level.level.index:
+                return -1
+            return 1
```
lemontree210, line 298: hm... not sure I understand why we're already returning `1` here without further checks

api/processors/services/group_builder.py
```diff
@@ -170,7 +164,7 @@ def _get_best_group_candidate(teacher_id: int) -> GroupCandidate | None:
         if len(group_candidates) == 0:
             return None
 
-        group_candidates.sort(key=lambda candidate: candidate.priority, reverse=True)
+        group_candidates.sort(key=functools.cmp_to_key(GroupBuilder._compare_groups_priority))
```
lemontree210, line 167: TIL

lemontree210 (commented):
Thank you for the PR and sorry for the delay! Overall, it looks very nice. We also need to deal with age ranges, though, IIRC.

api/processors/services/group_builder.py
Roxe322: I really think it was just copypaster mistake from below. Rewrote it a little bit, so it should be more clear now

api/processors/services/group_builder.py
Roxe322: Good catch. Reworked it - now we return 1 only if index1 is more than index2, but if they are same - we do compare by waiting time

api/models/auxil/constants.py
```diff
@@ -16,6 +16,9 @@ class LanguageLevelId(str, Enum):
     # no C2 at this school
 
 
+LANGUAGE_AND_LEVEL_ID_TO_INDEX = {level: index for index, level in enumerate(LanguageLevelId)}
```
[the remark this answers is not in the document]
Roxe322: Are we going to add any other levels? I thought no, so this order is fixed and shouldn't be changed. But I moved the constant and added `sorted`

api/models/language_and_level.py
```diff
@@ -1,10 +1,14 @@
...
 
-from api.models.auxil.constants import LANGUAGE_AND_LEVEL_ID_TO_INDEX, LanguageLevelId
+from api.models.auxil.constants import LanguageLevelId
 
 # Sometimes the languages and levels are needed separately, sometimes in combinations.
 # This means that we cannot create a "choices" field for language or level.
 
+"""Dictionary with names of levels ("A0", "A1" etc.) as values 
+and their position in the enumeration as keys."""
+LANGUAGE_LEVEL_ID_TO_INDEX = {level: index for index, level in enumerate(sorted(LanguageLevelId))}
```
lemontree210, line 10: nit: you have to put docstring below the constant for it to be picked up by IDE

lemontree210 (changes requested):
Thank you for the changes! Do you want to add age ranges in a separate PR? In this case, you can rename this one so it's clear that the algorithm is not complete yet.

api/models/language_and_level.py
Roxe322: TIL :)

Roxe322:
Let's discuss and implement age ranges separately, yep @lemontree210

lemontree210 (approved):
age ranges will be in a separate PR

Issue #437, opened 2024-02-23
fix (teachers populator) address owners' comments


lemontree210:
a follow-up on #406 that closed #405 

## Phones and Telegram usernames

In the legacy file, there is no separate "phone" column. In the "tg" column, teachers write either their phone number or their Telegram username (both are useful), or their full name on Telegram (like "tg: Ivan Ivanov"), thinking that one could find them by this name (wrong, hence not useful).

Therefore:
- if a cell contains a sequence of numbers (with or without separators between subsequences), try to parse it (I think `phonenumberslite` will be great for this). Sometimes these numbers don't contain country code (so we might want to try and parse them as Russian, Ukrainian or American phone numbers), sometimes they have `8` instead of `+7` for Russian numbers. If parsing is successful, store in `.phone` attribute of new `Teacher`.
- if a cell contains a sequence of alphanumeric characters starting with a `@` sign, store it in `telegram_username`. We'll try and clean up the source file so that all usernames get a `@` in front of them. Otherwise they will get confused with names and maybe emails (if some teachers put them there)

## Fix ID rule for silent teachers

If a teacher has status "waiting for response" and their ID is **lower** than 1300, it means that they have been in the legacy database for too long without responding. Skip those teachers, _not_ the ones with ID _greater_ than 1300.

PR #442, opened 2024-02-29
feature(populator): add student populator, address several problems

### LINKED ISSUE #408: Create populators for students

Auch-Auch:
Closes #408 and #437


django_webapps/scripts/db_population/parsers/common_parsers.py
```diff
@@ -88,11 +104,110 @@ def parse_telegram_name(tg_str: str) -> str | None:
...
+
+def find_any_phone_number(number_str: str) -> list[str]:
+    number_str = number_str.replace(" ", "")
+    regex = re.compile(r"[\+]?[(]?[0-9]{3}[)]?[-\s\.]?[0-9]{3}[-\s\.]?[0-9]{4,6}")
+    return re.findall(regex, number_str)
+
+
+def parse_phone_number(number_str: str) -> str | None:
+    number_str = number_str.strip().rstrip()
+    country_codes_to_test = ["", "+380", "+7", "+1"]
```
Auch-Auch, line 118: Note, this one does not matter, it can be anything, I think it is better to leave just one country code or use some +10000 just to distinguish it in the future

django_webapps/scripts/db_population/parsers/common_parsers.py
```diff
@@ -58,13 +61,26 @@ def parse_timezone(tz_str: str) -> datetime.tzinfo | None:
     return None
 
 
+def find_any_email(email_str: str) -> str | None:
+    regex = re.compile(
+        r"([-!#-'*+/-9=?A-Z^-~]+(\.[-!#-'*+/-9=?A-Z^-~]+)*|\"([]!#-[^-~ \t]|(\\[\t -~]))+\")@([-!#-'*+/-9=?A-Z^-~]+(\.[-!#-'*+/-9=?A-Z^-~]+)*|\[[\t -Z^-~]*])"  # noqa: E501
```
lemontree210, line 66: Django's `EmailField` could have different requirements. In the "dry" mode, the way it is now, we will not catch this error. I think we could [modify dry mode](https://adamj.eu/tech/2022/10/13/dry-run-mode-for-data-imports-in-django/) so that it involves Django models even if `dry` is `True`. This way we could avoid some other pitfalls we don't yet know about.

Otherwise our dry run will succeed, but then the actual writing to database might fail.

django_webapps/scripts/db_population/parsers/common_parsers.py
```diff
@@ -58,13 +61,26 @@ def parse_timezone(tz_str: str) -> datetime.tzinfo | None:
...
+    match = re.search(regex, email_str)
+    if match is not None:
+        return match.group(0)
+    return None
+
+
 def parse_email(email_str: str) -> str | None:
-    email_str = email_str.strip()
+
+    email_str_result: str | None = email_str.strip().rstrip()
```
lemontree210, line 76: `.strip().rstrip()` looks a bit odd (at least without a comment) :)

django_webapps/scripts/db_population/parsers/common_parsers.py
```diff
@@ -88,11 +104,110 @@ def parse_telegram_name(tg_str: str) -> str | None:
...
+        else:
+            range_to_fit = (5, int(re_groups[1]))
+        result.extend(find_best_fits(range_to_fit))
+    if len(result) > 0:
+        return result
+
+    # parse slots
+    # e.g. 11:00-17:00, 11-17, 11.17 and so on
+    regex_slots = re.compile(
+        r"(\d{1,2})[.\-:]{0,1}(\d{1,2}){0,1}[\-.](\d{1,2})[.\-:]{0,1}(\d{1,2}){0,1}"
```
lemontree210, line 192: add optional whitespaces? (`\s`)

django_webapps/scripts/db_population/parsers/student_parsers.py
```diff
@@ -0,0 +1,50 @@
...
+from api.models.choices.status.situational import StudentSituationalStatus
+
+RAW_PROJECT_STATUS_TO_PARSED: dict[
+    str, tuple[StudentSituationalStatus | None, StudentProjectStatus]
+] = {
+    "группа расформирована": (None, StudentProjectStatus.FINISHED),
+    "не ходит на занятия": (StudentSituationalStatus.NOT_ATTENDING, StudentProjectStatus.STUDYING),
+    "идут занятия": (None, StudentProjectStatus.STUDYING),
+    "нет ответа": (StudentSituationalStatus.NO_RESPONSE, StudentProjectStatus.NO_GROUP_YET),
+    "перестал отвечать": (StudentSituationalStatus.NO_RESPONSE, StudentProjectStatus.NO_GROUP_YET),
```
lemontree210, line 14: [Confirmed with the owners](https://t.me/c/1808422267/3115) that both cases mean that project status is "no group yet" 👍🏻

django_webapps/scripts/db_population/parsers/student_parsers.py
```diff
@@ -0,0 +1,50 @@
...
+    "вышел на связь, ожидает старта занятий": (
+        StudentSituationalStatus.AWAITING_START,
+        StudentProjectStatus.STUDYING,
+    ),
+    "уточнение расписания": (None, StudentProjectStatus.NO_GROUP_YET),
+    "жду ответа": (StudentSituationalStatus.NO_RESPONSE, StudentProjectStatus.NO_GROUP_YET),
```
lemontree210, line 20:
```suggestion
    "жду ответа": (None, StudentProjectStatus.NO_GROUP_YET),
```
we should set `.NO_RESPONSE` after a while. "жду ответа" is a neutral status (like, we wrote to the student but not enough time has passed to think that they're ignoring us).

django_webapps/scripts/db_population/populate_students.py
```diff
@@ -0,0 +1,164 @@
...
+    student_parsers,
+)
+from django_webapps.scripts.db_population.utils import (  # noqa: E402
+    get_args,
+    get_logger,
+    load_csv_data,
+)
+
+logger = get_logger("students.log")
+MIN_CID_WITH_NO_RESPONSE = 1300
```
lemontree210, line 34:
```suggestion
MIN_SID_WITH_NO_RESPONSE = 1300
```

django_webapps/scripts/db_population/populate_teachers.py
```diff
@@ -95,7 +95,7 @@ def _get_entity_data(self) -> TeacherData | None:
         tid = int(self._current_entity[self._column_to_id[self.id_name]])
         if parsed_status is None or (
             parsed_status[1] == TeacherSituationalStatus.NO_RESPONSE
-            and tid > MAX_TID_WITH_NO_RESPONSE
+            and tid < MIN_TID_WITH_NO_RESPONSE
         ):
             logger.info(
                 "Skipping teacher since there was no response from them or status is unclear"
```
lemontree210, line 101: same as with student

django_webapps/scripts/db_population/parsers/common_parsers.py
```diff
@@ -88,11 +104,110 @@ def parse_telegram_name(tg_str: str) -> str | None:
...
+            try:
+                parsed_number = phonenumbers.parse(number_to_parse)
+                break
+            except phonenumbers.NumberParseException:
+                pass
+
+    if parsed_number is None:
+        raise ValueError(f"Unable to parse phone number: {number_str}")
+
+    return phonenumbers.format_number(parsed_number, phonenumbers.PhoneNumberFormat.E164)
```
lemontree210, line 138: I think `phonenumbers` should have something like `.validate` too. A number could be parsed but be invalid. Like +1 314 123.

django_webapps/scripts/db_population/parsers/common_parsers.py
```diff
@@ -88,11 +104,110 @@ def parse_telegram_name(tg_str: str) -> str | None:
     if match_result is not None and len(match_result.group(0)) == len(tg_str):
         return f"@{match_result.group(0)}"
 
-    raise ValueError(f"tg: {tg_str} is not valid")
+    return None
+
+
+def find_any_phone_number(number_str: str) -> list[str]:
+    number_str = number_str.replace(" ", "")
+    regex = re.compile(r"[\+]?[(]?[0-9]{3}[)]?[-\s\.]?[0-9]{3}[-\s\.]?[0-9]{4,6}")
```
lemontree210, line 112: hm... I'm not sure about numbers in curly braces, they may make the regex way too strict. Also, I think you meant `{1,3}` in the first pair of braces (country code?)

I'd rather be more lenient here and fail on the parsing stage

lemontree210 (changes requested):
Thanks for the tremendous effort!

django_webapps/scripts/db_population/parsers/common_parsers.py
Auch-Auch: I handle it one line below

django_webapps/scripts/db_population/parsers/common_parsers.py
Auch-Auch: It removes whitespaces from the right and left, yes, the API kind of strange )

django_webapps/scripts/db_population/populate_teachers.py
Auch-Auch: Do not get it

django_webapps/scripts/db_population/parsers/common_parsers.py
Auch-Auch: I tried that but there were tons of failed numbers so I decided to find just any number and try, this one more or less provide a valid number that we can then try to insert in DB

django_webapps/scripts/db_population/parsers/common_parsers.py
Auch-Auch: Yes, you are right, will do a refactoring, but here it is enough to just use validate_email for every string since django uses the same lib

django_webapps/scripts/db_population/parsers/common_parsers.py
Auch-Auch: The validation is working only if we provide a country code but we can not do it in advance

django_webapps/scripts/db_population/parsers/common_parsers.py
```diff
@@ -88,11 +106,119 @@ def parse_telegram_name(tg_str: str) -> str | None:
...
+    return re.findall(regex, number_str)
+
+
+def parse_phone_number(number_str: str) -> str | None:
+    number_str = number_str.replace(" ", "")
+    country_codes_to_test = [("", None)] + [
+        (f"+{code}", region[0])
+        for code, region in phonenumbers.COUNTRY_CODE_TO_REGION_CODE.items()
```
Auch-Auch, line 122: Note, even like that it will not parse all and we will have like ~400 invalid numbers, not sure how to parse them manually

django_webapps/scripts/db_population/parsers/common_parsers.py
lemontree210: IIRC you can set a country, and then `phonenumbers` will try and validate the number as a regional phone number for that country. We can try several most common countries: Ukraine, Russia, UK, Ireland, US, Germany, France, Spain

django_webapps/scripts/db_population/parsers/common_parsers.py
lemontree210: Yes I see that `\+` is optional. But is the exact `{3}` correct? I have some doubts there.
Also, is it really necessary to use brackets before question mark when there's only one option? I don't think I have used it in my regular expressions with question marks, just did `\s?` and I think it worked :) Of course, when there are several options, you need brackets

django_webapps/scripts/db_population/parsers/common_parsers.py
lemontree210: Can I look at these failing numbers? Maybe I could come up with some idea

django_webapps/scripts/db_population/parsers/common_parsers.py
Auch-Auch: Yes, I did that already for all coutnries

django_webapps/scripts/db_population/parsers/common_parsers.py
Auch-Auch: Ah, I see, this one in particular generated by ChatGPT, string {3} in "()" brackets  clause which is correct, I think, the other one also seems to be if we take into account that minimum number of digits in the number is 7, regarding brackets, yes, we are not needed, not sure why he decided to add it

django_webapps/scripts/db_population/parsers/common_parsers.py
lemontree210: I would still think that at least some note in comment field (phone number passed or failed `phonenumbers`' validation) would be very useful for the coordinators/admins.

django_webapps/scripts/db_population/parsers/common_parsers.py
Auch-Auch: I have tried different ones but more or less same result, let's stick with the last one

lemontree210: I think the second `{3}` could be correct (area code), although in some European countries area codes only have two digits IIRC, but the first `{3}` surely is wrong, isn't it? Country code can have `{1,3}` digits

lemontree210: Also, `{4,6}` seems wrong as well because for example in Berlin there are numbers consisting of 8 digits **after** the area code

lemontree210 (approved):
LGTM, but please send me another log before merging so we can look at failing phone numbers again

lemontree210 (changes requested):
Okay, let's just get the `00...` phone numbers to be parsed correctly and then merge. I thought `phonenumbers` was able to do it, but if not, we can just replace leading `00` with a `+` and then parse/validate

lemontree210 (approved):
thank you!

PR #460, opened 2024-03-16
Add baseline for group population

### LINKED ISSUE #454: Create populators for groups
The main problem to solve to find students, teachers and coordinators for groups by their old IDs (SID, TID and CID respectively) so they can be "attached" to the group being created.

@Auch-Auch maybe you will decide to make some changes to their populators as well (e.g. introduce some file that matches old IDs to new ones).

Auch-Auch:
Closes #454


api/models/shared_abstract/group_or_person.py
```diff
@@ -9,6 +9,7 @@
 class GroupOrPerson(models.Model):
     """Model holding attributes that are common for every person and group."""
 
+    legacy_id = models.IntegerField(null=True)
```
Auch-Auch, line 12: Not sure what is more safe, to have unique constraint or not, for now I think it is okay

django_webapps/scripts/db_population/parsers/group_parsers.py
```diff
@@ -0,0 +1,44 @@
...
+RAW_PROJECT_STATUS_TO_PARSED: dict[str, GroupProjectStatus] = {
+    "группа расформирована": GroupProjectStatus.FINISHED,
+    "уровень завершен": GroupProjectStatus.FINISHED,
+    "идет набор": GroupProjectStatus.WORKING,
+    "нет набор": GroupProjectStatus.AWAITING_START,
+}
+
+
+LANGUAGES_MAPPING: dict[str, str] = {
+    "русский": "Russian",
```
Auch-Auch, line 12: These are all in the old base but in the new one we do not have all of them

api/models/shared_abstract/group_or_person.py
lemontree210: Here's the thing: one person can have multiple IDs: e.g. one as teacher, one as coordinator. That's why we switched to `PersonalInfo` in the first place: to create a unique ID per person, not per role of person. So there cannot be just one `.legacy_id`.

That's why I initially suggested creating `.legacy_sid` in `Student`, `.legacy_tid` in `Teacher` etc. And then I forgot my own motivation when we talked about moving the legacy ID into the abstract model.

Auch-Auch: So you want to create a single personal info for one "person", not sure it is possible, how we map between different roles?

Auch-Auch: Any way we will have multiple personal infos for a single person

lemontree210: > Any way we will have multiple personal infos for a single person

No, it's the other way round 😸 

Each human individual is described by only one `PersonalInfo` object. This object is then used as a primary key for `Person` abstract class, e.g. to create `Student`s, `Coordinator`s and teachers. The [way it's created](https://github.com/samanthasgroup/django-webapps/blob/master/api/models/shared_abstract/person.py#L13C21-L18) ensures that for each `PersonalInfo` there can be only one `Student` stored in `.as_student` attribute, only one `Coordinator` as `.as_coordinator` etc.

lemontree210: Actually, this means that we should tweak the `BasePopulator` and use `.get_or_create()` instead of `.create` in [this method](https://github.com/samanthasgroup/django-webapps/blob/master/django_webapps/scripts/db_population/base_populator.py#L115-L130) (maybe it could be renamed too). I guess I should create a separate issue for this.

Auch-Auch: I meant exactly that, how do you identify a duplicate?

Auch-Auch: If there any rule to use exactly one email or name?

Auch-Auch: It does make sense to not create a separate Personal info if email is duplicated

Auch-Auch: And we can filter by unique emails from the input CSV, we just need clarification from the owners that emails are unique (usually at least)

lemontree210: > is there any rule to use exactly one email or name?

The combination of `first_name`, `last_name` and `email` must be unique.

Since people can e.g. register their children, `email` alone is not good for uniqueness test.

api/models/coordinator.py
```diff
@@ -37,6 +37,7 @@ def filter_active(self) -> "CoordinatorQuerySet":
 class Coordinator(Person):
     """Model for a coordinator."""
 
+    legacy_cid = models.IntegerField(null=True)  # from the old database
```
lemontree210, line 40: I think that this comment can be put into `help_text`. It will be used both as a comment here and as a hint in Django admin

lemontree210 (changes requested):
To be clear: have we agreed that changes to `create_personal_info` will be made in a separate PR?

I'm selecting "Request changes" for the review because some tests are failing

lemontree210 (approved):

Auch-Auch:
Do I merge?

lemontree210:
> Do I merge?

absolutely!

PR #470, opened 2024-03-23
Add create or get personal info for teacher, add readme for scripts

### LINKED ISSUE #464: Allow existing `PersonalInfo` to be used when a new person is created
Suppose we have moved a coordinator from legacy DB to the new one. Then we find a teacher in legacy DB that has same personal data (full name and email). We cannot create another `PersonalInfo` record, but we should instead us the already created record and link the new `Teacher` to it.

If, however, we find another coordinator with same personal data, it must not be accepted. But the checks performed during the dry run will prevent this from happening anyway. Maybe we should add a safety net anyway, it's up to you @Auch-Auch .

Auch-Auch:
Closes #464

### DIFF
--- a/django_webapps/scripts/db_population/README
+++ b/django_webapps/scripts/db_population/README
@@ -1,0 +1,10 @@
+The scripts should be run in the following order: 
+
+- populate_coordinators.py with coordinators database exported in CSV
+- populate_teachers.py with teacher database (CSV)
+- populate_students.py with students databse (CSV)
+- populate_groups.py with groups database (CSV)
+
+To run all the scripts: 
+
+> sh populate_all.sh coord_base.csv teacher_base.csv student_base.csv group_base.csv
--- a/django_webapps/scripts/db_population/populate_teachers.py
+++ b/django_webapps/scripts/db_population/populate_teachers.py
@@ -15,6 +15,7 @@ from django_stubs_ext import QuerySetAny  # noqa: E402
 from api.models.age_range import AgeRange  # noqa: E402
 from api.models.choices.status.project import TeacherProjectStatus  # noqa: E402
 from api.models.choices.status.situational import TeacherSituationalStatus  # noqa: E402
+from api.models.personal_info import PersonalInfo  # noqa: E402
 from api.models.teacher import Teacher  # noqa: E402
 from django_webapps.scripts.db_population.base_populator import (  # noqa: E402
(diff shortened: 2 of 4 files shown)


django_webapps/scripts/db_population/README
```diff
@@ -0,0 +1,10 @@
+The scripts should be run in the following order: 
+
+- populate_coordinators.py with coordinators database exported in CSV
+- populate_teachers.py with teacher database (CSV)
+- populate_students.py with students databse (CSV)
+- populate_groups.py with groups database (CSV)
+
+To run all the scripts: 
+
+> sh populate.sh coord_base.csv teacher_base.csv student_base.csv group_base.csv
```
lemontree210, line 10: `populate_all.sh`?

django_webapps/scripts/db_population/populate_teachers.py
```diff
@@ -216,6 +217,17 @@ def _create_age_ranges(self, is_teaching_children: bool) -> QuerySetAny[AgeRange
...
+    def _create_or_get_personal_info(self, entity_data: TeacherData) -> None | PersonalInfo:
+        already_existing_one = PersonalInfo.objects.filter(
+            first_name=entity_data.first_name,
+            last_name=entity_data.last_name,
+            email=entity_data.email,
+        ).first()
+        if already_existing_one:
+            logger.info("Personal info already exists")
+            return already_existing_one
+        return super()._create_personal_info(entity_data)
```
lemontree210, line 229: maybe simply `get_or_create()`? ([docs](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#get-or-create))

lemontree210 (commented):
Maybe the PR is not "for teacher" but for any Person

django_webapps/scripts/db_population/populate_teachers.py
Auch-Auch: As far as I see from the description, it is for a different use case, this one is simple enough I think

lemontree210: Well, alright :) It may not be safe for concurrent requests, but we will have none in these scripts

lemontree210 (approved):

lemontree210:
@Auch-Auch these force pushes confuse me a bit :) I see there's still one commit and then it turns out you fixed my comments)

PR #475, opened 2024-03-29
fix: add handler for GeocoderUnavailable error

Auch-Auch:
Sometimes geocoder is unavailable for some reason, some strings are not valid for him, apparently


lemontree210:
> some strings are not valid for him, apparently

could you find any specifics?

Auch-Auch:
GM+, for example

lemontree210:
> GM+, for example

GM+ doesn't look like a valid string for me either :)

Auch-Auch:
Yes, but it used to worked so API changed apparently

Auch-Auch:
I mean in this case it should not throw that exception but return None or some other exception since not city is found

lemontree210 (approved):

PR #482, opened 2024-04-02
feat: add line breakers at different columns names

### LINKED ISSUE #415: Admin site: make selected columns in tables narrower
Make selected columns narrower to avoid excessive horizontal scrolling.

- All tables: add line break in `SITUATIONAL STATUS` (to make "status" move to the 2nd line). Already done for `Coordinator` in #377 
- ~Coordinators~ (done in #377), ~Students, Teachers~ (not created yet): make `COMMENT` narrower
- Groups: add line break to header `NUMBER OF STUDENTS`

Auch-Auch:
Closes #415

### DIFF
--- a/api/admin/group.py
+++ b/api/admin/group.py
@@ -60,8 +60,8 @@ class GroupAdmin(VersionAdmin):
         "get_start_date",
         "get_end_date",
         "get_schedule",
-        "project_status",
-        "situational_status",
+        "get_project_status",
+        "get_situational_status",
         "get_status_since",
         "staff_only",
     )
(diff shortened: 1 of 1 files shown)


Auch-Auch:
Note: there is no student or teacher admin pages

lemontree210:
> Note: there is no student or teacher admin pages

Yes, the idea was that the owners will manage coordinators in Django admin and teachers/students in ToolJet to avoid double products. I'm investigating what they prefer and whether it makes sense to actually create tables for students and teachers in Django admin. If we decide to do so, this should be less time consuming than creating coordinators' tables from scratch.

PR #483, opened 2024-04-02
Override action template and group name in coordinator page

### LINKED ISSUE #413: Admin site: Cosmetic changes to actions interface in table of coordinators
After #268 , owners requested some cosmetic changes to the "actions" interface.

- in the "Group" drop-down list, group names are too verbose. Leave only group ID, language/level, coordinator(s). 
For example: `1: English A2, (coords 2: Johns Smith, 3: Peter Powell)`*
- rename "Go" button into "Apply selected action"
- change order: 1st line - Group, 2nd line - Action, 3rd line - button

\* representation of a person as `id: first name last name` is a default `__str__()` representation of `Person` now, AFAIR

Auch-Auch:
Closes #413

>[!IMPORTANT] 
> `settings.py` will need to be updated on server after merging

### DIFF
--- a/api/admin/coordinator.py
+++ b/api/admin/coordinator.py
@@ -56,11 +56,8 @@ class BaseCoordinatorGroupInline(
 class NameGroupChoiceField(forms.ModelChoiceField):
 
     def label_from_instance(self, obj: Any) -> str:
-        coordinators = ",".join([str(c) for c in obj.coordinators.all()]).rstrip().strip()
-        return (
-            f"Group {obj.pk}, {obj.language_and_level} (coords: {obj.coordinators.count()}, "
-            f"coordinators: {coordinators})"
-        )
+        coordinators = ", ".join([str(c) for c in obj.coordinators.all()]).rstrip().strip()
+        return f"Group {obj.pk}, {obj.language_and_level} (coordinators: {coordinators})"
 
 
 class GroupActionForm(ActionForm):
--- a/api/admin/coordinator.py
+++ b/api/admin/coordinator.py
@@ -53,8 +53,21 @@ class BaseCoordinatorGroupInline(
     max_num = 0
 
 
+class NameGroupChoiceField(forms.ModelChoiceField):
+
+    def label_from_instance(self, obj: Any) -> str:
+        coordinators = ",".join([str(c) for c in obj.coordinators.all()]).rstrip().strip()
+        return (
+            f"Group {obj.pk}, {obj.language_and_level} (coords: {obj.coordinators.count()}, "
+            f"coordinators: {coordinators})"
+        )
+
+
 class GroupActionForm(ActionForm):
-    group = forms.ModelChoiceField(queryset=Group.objects.all(), required=False)
+    group = NameGroupChoiceField(
+        queryset=Group.objects.all(),
+        required=False,
+    )
 
 
 class CoordinatorActiveGroupsInline(BaseCoordinatorGroupInline):
--- a/api/admin/coordinator.py
+++ b/api/admin/coordinator.py
@@ -57,7 +57,7 @@ class NameGroupChoiceField(forms.ModelChoiceField):
 
     def label_from_instance(self, obj: Any) -> str:
         coordinators = ", ".join([str(c) for c in obj.coordinators.all()]).rstrip().strip()
-        return f"Group {obj.pk}, {obj.language_and_level} (coordinators: {coordinators})"
+        return f"Group {obj.pk}, {obj.language_and_level} (coords: {coordinators})"
 
 
 class GroupActionForm(ActionForm):
--- a/django_webapps/settings_sample.py
+++ b/django_webapps/settings_sample.py
@@ -10,6 +10,7 @@ https://docs.djangoproject.com/en/4.1/topics/settings/
 For the full list of settings and their values, see
 https://docs.djangoproject.com/en/4.1/ref/settings/
 """
+import os
 from pathlib import Path
 
 import django_stubs_ext
@@ -61,7 +62,7 @@ ROOT_URLCONF = "django_webapps.urls"
 TEMPLATES = [
     {
         "BACKEND": "django.template.backends.django.DjangoTemplates",
-        "DIRS": [],
+        "DIRS": [os.path.join(BASE_DIR, "templates/")],  # noqa: PTH118
         "APP_DIRS": True,
         "OPTIONS": {
             "context_processors": [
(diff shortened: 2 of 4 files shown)


Auch-Auch:
Looks like this for all pages: 
image

lemontree210:
> Looks like this for all pages (for other pages without group): image

Thank you, it looks great! However, they [would like to](https://t.me/c/1808422267/2792) have group, then action, then button :)

templates/admin/actions.html
```diff
@@ -0,0 +1,28 @@
...
+  <div class="actions">
+    {% for field in action_form %}
+      {% if field %}
+        <div class="action-field">{{ field.label }} {{ field }}</div>
+      {% endif %}
+    {% endfor %}
+
+    <div class="button-action">
+      <button type="submit" class="button small" title="{% trans "Run the selected action" %}" name="index" value="{{ action_index|default:0 }}">
+          {% trans "Apply selected action" %}
```
lemontree210, line 13: Oh I see it's "action" (without "s") already, so maybe the screenshot was outdated

templates/admin/actions.html
```diff
@@ -0,0 +1,28 @@
...
+{% if action_form %}
+  <div class="actions">
+    {% for field in action_form %}
+      {% if field %}
+        <div class="action-field">{{ field.label }} {{ field }}</div>
+      {% endif %}
+    {% endfor %}
+
+    <div class="button-action">
+      <button type="submit" class="button small" title="{% trans "Run the selected action" %}" name="index" value="{{ action_index|default:0 }}">
```
lemontree210, line 12: how does this button get the index? is it something that is built in into Django admin functionality?

api/admin/coordinator.py
```diff
@@ -53,8 +53,21 @@ class BaseCoordinatorGroupInline(
     max_num = 0
 
 
+class NameGroupChoiceField(forms.ModelChoiceField):
+
+    def label_from_instance(self, obj: Any) -> str:
+        coordinators = ",".join([str(c) for c in obj.coordinators.all()]).rstrip().strip()
+        return (
+            f"Group {obj.pk}, {obj.language_and_level} (coords: {obj.coordinators.count()}, "
```
lemontree210, line 61:
```suggestion
            f"Group {obj.pk}, {obj.language_and_level} ({obj.coordinators.count()} students, "
```
maybe you meant students?

api/admin/coordinator.py
```diff
@@ -53,8 +53,21 @@ class BaseCoordinatorGroupInline(
     max_num = 0
 
 
+class NameGroupChoiceField(forms.ModelChoiceField):
+
+    def label_from_instance(self, obj: Any) -> str:
```
lemontree210, line 58: is `Group`'s `__str__` method not good enough?

templates/admin/actions.html
Auch-Auch: I think it is from the action form and passed here within context but it will be always None or 0 since we do not have query params or any default action specified

Auch-Auch: As far as I know it is used only as default value

api/admin/coordinator.py
Auch-Auch: Then we will change it for everything, do we want it?

api/admin/coordinator.py
Auch-Auch: Wait, why students?

templates/admin/actions.html
```diff
@@ -0,0 +1,28 @@
+{% load i18n %}
+
+{% if action_form %}
+  <div class="actions">
+    {% for field in action_form reversed %}
```
Auch-Auch, line 5: Note that action will be always at the bottom

api/admin/coordinator.py
lemontree210: Dunno, but you repeat coordinators twice, first their number, then their names, I just didn't think it was needed. OTOH, it might make sense to just show how many students are in the group.
Anyway, I may be confusing you needlessly

api/admin/coordinator.py
lemontree210: But I think `Group.__str__()` method already gives [decent enough information](https://github.com/samanthasgroup/django-webapps/blob/master/api/models/group.py#L119-L125) as it is, does it not?

Auch-Auch: image

templates/admin/actions.html
lemontree210: Hm... I just don't like having code about which I don't understand how it works... I see similar code snippets on SO and GitHub, but there is no explanation.

api/admin/coordinator.py
lemontree210: > image

Oh, right, things get erased from my memory. They requested it to be less verbose. Well, then "2: " is coordinator's ID :) So it should be something like
```py
",".join([f"{coordinator.id}: {coordinator.name}" for coordinator in obj.coordinators])
```

api/admin/coordinator.py
lemontree210: yep, [I am](https://github.com/samanthasgroup/django-webapps/pull/483#discussion_r1552484422)

closing this thread

templates/admin/actions.html
lemontree210: Not sure I get this one without seeing it :)

lemontree210 (changes requested):

api/admin/coordinator.py
lemontree210: Did it in https://github.com/samanthasgroup/django-webapps/pull/483/commits/92c2a8e77dde651d269770e1a66547b57f6a7192

django_webapps/settings_sample.py
```diff
@@ -61,7 +62,7 @@
 TEMPLATES = [
     {
         "BACKEND": "django.template.backends.django.DjangoTemplates",
-        "DIRS": [],
+        "DIRS": [os.path.join(BASE_DIR, "templates/")],  # noqa: PTH118
```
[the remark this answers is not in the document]
lemontree210: Did it in https://github.com/samanthasgroup/django-webapps/pull/483/commits/8124553aa2f361290e0816c8c73b33dfacab7759

templates/admin/actions.html
lemontree210: > I think it is from the action form and passed here within context but it will be always None or 0 since we do not have query params or any default action specified

Yes, you're right:

> The `value` attribute holds the data that will be sent when the form is submitted. In this case, it's using the `action_index` variable from the Django context. If `action_index` is not provided in the context, it defaults to `0`. This value helps identify which action was selected when the form is processed on the server side.
> 
> The `action_index` is a variable that should be passed to the template context by the Django view that renders this template. It's not defined within the template itself; it's part of the context data provided by the view. If action_index is not included in the context, the default filter sets it to `0` as a fallback. This ensures that the form can still be submitted without errors even if `action_index` is missing.

lemontree210 (approved):
Now everything LGTM

Issue #506, opened 2024-04-26
[COMPOUND] Address owner's comments on table of coordinators


lemontree210:
- [ ] #505

PR #512, opened 2024-04-28
Make `.personal_info.id` clickable in table of `Coordinator`s

### LINKED ISSUE #505: Admin site: `Coordinator.personal_info.full_name` clickable (not only ID)
UPD making the whole row clickable seems to difficult now. Changing the task to just make `.personal_info.full_name` clickable in addition to `.personal_info.id`

SivikGosh:
closes #505 

Initially this was an attempt to make the whole row clickable, but it's too difficult and may not be worth the effort for now


api/migrations/0004_alter_coordinator_legacy_cid_alter_group_legacy_gid_and_more.py
lemontree210, line 1: We regenerate the initial migration instead of creating a new one. This will be the case at least until the launch of MVP. We even [have a command](https://github.com/samanthasgroup/django-webapps/blob/master/Makefile#L14) in our Makefile for that :)

Also, I think we should move this to a separate PR, along with deletion of `.legacy_id` in `GroupOrPerson`. We replaced it with specific legacy IDs. So I guess we should remove `.legacy_id` in `GroupOrPerson`  and create `.legacy_gid` in `Group`. @Auch-Auch maybe you could do this one :)

api/admin/coordinator.py
```diff
@@ -154,6 +154,19 @@ class CoordinatorAdmin(VersionAdmin):
...
+        "get_personal_info_full_name",
+        "get_additional_skills_comment",
+        "get_project_status",
+        "get_situational_status",
+        "get_status_since",
+        "active_groups_count",
+        "get_communication_language_mode",
+        "mentor",
+        "get_comment",
+    )
```
lemontree210, line 168: I understand the solution, but the idea was to make the whole row clickable, not text-containing cells in it. 

It's a more difficult task, I know. It probably should even include the highlighting of the row the mouse cursor is hovering over.

If it's not doable at all, we should only leave ID and full name clickable, otherwise it would look confusing

lemontree210 (changes requested):

api/migrations/0004_alter_coordinator_legacy_cid_alter_group_legacy_gid_and_more.py
Auch-Auch: Apparently initial migration was not merged in the last PR

api/admin/coordinator.py
SivikGosh: I got it. I'll look how could be realize it. :)

templates/admin/change_list.html
```diff
@@ -0,0 +1,93 @@
+{% extends "admin/base_site.html" %}
```
Auch-Auch, line 1: @SivikGosh @lemontree210  I believe there should be a better solution otherwise I would put this task to the backlog since it requires too much effort. I think this violates the core idea of django which is to have everything out of the box and reuse pre-created/default components as much as possible since this is the only reason django exists (to save time and effort while ensuring quality) and if such a small task requires so many changes in these components I do not think it is worth it.

api/migrations/0004_alter_coordinator_legacy_cid_alter_group_legacy_gid_and_more.py
```diff
@@ -0,0 +1,33 @@
...
+
+    dependencies = [
+        ("api", "0003_fake_data"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="coordinator",
+            name="legacy_cid",
+            field=models.IntegerField(help_text="Coordinator id from the old database", null=True),
```
Auch-Auch, line 16: Please, rebase your branch, remove this file and do: `make recreate-first-migration`

Auch-Auch (changes requested):

templates/admin/change_list.html
```diff
@@ -0,0 +1,93 @@
...
+        {% block date_hierarchy %}{% if cl.date_hierarchy %}{% date_hierarchy cl %}{% endif %}{% endblock %}
+
+        <form id="changelist-form" method="post"{% if cl.formset and cl.formset.is_multipart %} enctype="multipart/form-data"{% endif %} novalidate>{% csrf_token %}
+        {% if cl.formset %}
+          <div>{{ cl.formset.management_form }}</div>
+        {% endif %}
+
+        {% block result_list %}
+            {% if action_form and actions_on_top and cl.show_admin_actions %}{% admin_actions %}{% endif %}
+            {% clickable_result_list cl %}
```
Auch-Auch, line 68: I was not able to compile it, please, make sure you have added changes from settings.py or any other changes file

Auch-Auch:
@lemontree210 In general I think we should not focus on such tasks at all. We can do the following: whatever cosmetic changes (which do not affect target business logic) requested by the owners should be put at one task under list of sub-tasks and do it only if it is possible to implement it via standard django admin interface (meaning without much effort) otherwise we will never release

lemontree210:
> In general I think we should not focus on such tasks at all. 

I see your point. I outlined my rationale [here](https://t.me/c/1808422267/3280), and here I can just repeat that, given my concerns over ToolJet's stability, these tasks become not just 'cosmetic' but potentially very important UX features

templates/admin/change_list.html
lemontree210: > I think this violates the core idea of django which is to have everything out of the box and reuse pre-created/default components as much as possible since this is the only reason django exists.

I can only agree to a certain extent. Yes, Django strives to provide _basic/standard_ components out of the box, but I haven't heard of users being discouraged from e.g. editing templates because this would violate Django's philosophy. I think even the official docs contain info on how to customize admin site by changing its templates.

lemontree210:
> Whatever cosmetic changes (which do not affect target business logic) requested by the owners should be put at one task under list of sub-tasks 

This is exactly what I did, I think :) See #506 

> ... and do it only if it is possible to implement it via standard django admin interface (meaning without much effort) 

Here's where I disagree: when picking Django, we never made decisions to stick only to ready-made things 

> otherwise we will never release. 

I understand your concern, but see, our colleagues doing ToolJet are still working on the owners' comments, some of which concern basic functionality, and our bot 'division' is behind the schedule as well, because it consists only of me, while all other potential developers that could help with the bot are unavailable at the moment. So by asking @SivikGosh to do these features, I

1. Do not delay the release of MVP
2. Try to move towards potentially replacing ToolJet. All of these 'cosmetic' changes will be applicable to any other model in Django admin: students, teachers, and groups.

> In any possible case it will be batter then EXCEL sheets

Agreed, but being just "better than Google Sheets" is not enough :(

lemontree210:
@Auch-Auch @SivikGosh all in all, in this concrete task I would say that making `full_name` clickable is already more intuitive for the owners. Maybe we should do the following:

- stick with making `full_name` clickable for now and save the template-changing ideas for later
- see if the other tasks in #506 require changes to the template
- if they do, we can try and apply changes for clickable rows in a later PR

SivikGosh:
Maybe we should make coordinator's pages, outer from admin panel? With Django templates and personal path.

lemontree210:
> @lemontree210 @Auch-Auch, maybe we should make coordinator's pages, outer from admin panel? With Django templates and personal path.

Sorry for the delayed reply! I'm not sure what you mean here, sorry :)

templates/admin/change_list.html
lemontree210: I see no clickable rows, unfortunately.

Actually, I think that some JS should do the trick. But let's just forget it all for now, make the full name clickable and leave the task for clickable rows in backlog

api/migrations/0004_alter_coordinator_legacy_cid_alter_group_legacy_gid_and_more.py
lemontree210: I did this now

lemontree210:
> @Auch-Auch @SivikGosh all in all, in this concrete task I would say that making `full_name` clickable is already more intuitive for the owners. Maybe we should do the following:
> * stick with making `full_name` clickable for now and save the template-changing ideas for later

I implemented this for now. @Auch-Auch please approve to observe formalities :)

PR #517, opened 2024-05-01
fix: fix mypy errors, remove legacy id

Auch-Auch:
See #516 and https://github.com/samanthasgroup/django-webapps/pull/512#discussion_r1582367700


lemontree210 (approved):
LGTM

PR #548, opened 2024-06-12
Coordinators' table: add sorting by other columns (make cells in header row clickable)

### LINKED ISSUE #547: Add sorting by other columns (only sorting by ID is available now), making column headers clickable to enable sorting
https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display

lemontree210:
closes #547

### DIFF
--- a/api/admin/coordinator.py
+++ b/api/admin/coordinator.py
@@ -162,45 +162,48 @@ class CoordinatorAdmin(VersionAdmin):
     class Media:
         js = ("admin/js/sticky-scroll-bar.js",)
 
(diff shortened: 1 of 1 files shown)


Auch-Auch (approved):
LGTM, maybe we should change the PR name? It is hard to understand without opening the issue.

lemontree210:
> LGTM, maybe we should change the PR name? It is hard to understand without opening the issue.

done (sorry for an incredibly long response, I had messed up email notifications)

